PR #1, opened 2023-01-23
Internal: Initial commit

christianvuerings:
# FAQ

## Why pnpm?

https://pnpm.io/motivation

* Single copy of dependencies
* Save disk space + way faster installations 
* Workspace support: https://pnpm.io/workspaces 

## Why changesets

https://github.com/changesets/changesets/blob/main/docs/intro-to-using-changesets.md


packages/syntax-core/src/index.tsx
```diff
@@ -0,0 +1 @@
+export { Divider, type DividerProps } from "./Divider";
```
christianvuerings, line 1: @christianvuerings remove DividerProps

PR #40, opened 2023-03-18
MiniActionCard

jliotta:
This PR adds the MiniActionCard. Right now I'm keeping it fairly limited and just a container component like `Card`. We can iterate on it in the future but these changes are needed for the student dashboard mvp

Screen Shot 2023-03-17 at 7 56 46 PM

Screen Shot 2023-03-17 at 8 05 08 PM


packages/syntax-core/src/MiniActionCard/MiniActionCard.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+        style={{
+          display: "flex",
+          flexDirection: "column",
+          alignItems: "center",
+          justifyContent: "center",
+          height: "100%",
+          gap: "24px",
+        }}
+      >
+        {/* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment */}
```
jliotta, line 30: I can't figure out what's wrong with this line or why eslint was throwing this error

jliotta: Also not sure what's wrong with Vercel. I don't have access to it

christianvuerings: @jliotta Did you not see the following warning in your editor? 

> Using `` could result in slower LCP and higher bandwidth. Consider using `<Image />` from `next/image` to automatically optimize images. This may incur additional usage or cost from your provider. See: https://nextjs.org/docs/messages/no-img-elementeslint[@next/next/no-img-element](https://nextjs.org/docs/messages/no-img-element)

image

Issue here is that we have the `next` ESLint rules enable for all or `syntax-core`. That might make sense for most of our components but not our stories. So I disabled that.

christianvuerings: @jliotta As for vercel access, I just invited you

The error was the following:
> @cambly/syntax-core:build: src/index.tsx(4,28): error TS2307: Cannot find module './MiniActionCard' or its corresponding type declarations.
17:25:21.579

And the issue was a wrong import:
image

Changing that to `MiniCard/MiniCard` fixes the issue:
image


I also added `pnpm run build` to GitHub CI so folks will be able to see these errors there instead:
Screenshot 2023-03-20 at 9 10 00 AM

christianvuerings (approved):
@jliotta no blockers but would be good to see the comments addressed

packages/syntax-core/src/MiniActionCard/MiniActionCard.tsx
```diff
@@ -0,0 +1,15 @@
...
+ * MiniActionCard is component that alerts users to a call to action.
+ */
+const MiniActionCard = ({
+  children,
+}: {
+  /**
+   * The child components to render within MiniActionCard.
+   */
+  children: JSX.Element;
+}): JSX.Element => <div className={styles.miniActionCard}>{children}</div>;
```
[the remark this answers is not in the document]
jliotta: @christianvuerings The padding is different and we also set the height and width specifically for `MiniActionCard`. Would we still want to use `Card` in this case? Would we have to find a way to pass in this custom css without also exposing that ability to users of `Card`?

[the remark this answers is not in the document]
christianvuerings: @jliotta makes sense - let's keep them separate for now

PR #42, opened 2023-03-21
Button / ButtonGroup: add components

christianvuerings:
# Changes

* Add `Button` and `ButtonGroup` components
* Convert Stylelint configuration to a JavaScript file

# Notes

* We don't yet have Typography in Syntax but will use that within Button once we move that component over.

# Screenshots

## Button
Button-Docs-⋅-Storybook

## ButtonGroup
ButtonGroup-Docs-⋅-Storybook


ajmooo:
Can you rename "danger" to "destructive" to match the destructive color name in the system? I'll make sure my figma is updated to reflect this change.

ajmooo (approved):

christianvuerings:
> Can you rename "danger" to "destructive" to match the destructive color name in the system? I'll make sure my figma is updated to reflect this change.

@ajmooo Yes! I converted all `danger` references to `destructive`

packages/syntax-core/src/Button/Button.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+    },
+    fullWidth: {
+      control: "boolean",
+    },
+    onClick: { action: "clicked" },
+  },
+  tags: ["autodocs"],
+} as Meta<typeof Button>;
+
+export const Default: StoryObj<typeof Button> = {
```
somethiiing, line 46: should there be a story using end/start icon?

packages/syntax-core/src/Button/Button.module.css
```diff
@@ -0,0 +1,143 @@
...
+
+.lg {
+  min-width: 88px;
+  height: 48px;
+  padding-left: 20px;
+  padding-right: 20px;
+  border-radius: 16px;
+}
+
+.smIcon {
```
somethiiing, line 78: if we're inlining the svgs, will we need this stuff still because we should be able to tweak these directly?

packages/syntax-core/src/Button/Button.module.css
```diff
@@ -0,0 +1,143 @@
...
+}
+
+.fullWidth {
+  width: 100%;
+}
+
+.sm {
+  min-width: 52px;
+  height: 32px;
+  padding-left: 12px;
```
somethiiing, line 57: would these padding-left/rights be better off as `padding: 0 12px`? i don't see that we're explicitly setting padding top/bottom anywhere

somethiiing (approved):
looks good, left some minor comments

packages/syntax-core/src/Button/Button.module.css
christianvuerings: @somethiiing Yes, updated

packages/syntax-core/src/Button/Button.module.css
christianvuerings: @somethiiing For now we are still using Material UI icons. I noticed that they have more than 2000 icons in 5 different variations so it would be a lot of work to maintain those

packages/syntax-core/src/Button/Button.stories.tsx
christianvuerings: @somethiiing Yes, updated

PR #61, opened 2023-03-24
Typography & Heading: add components to Syntax

christianvuerings:
# Changes

* Adds `Typography` and `Heading` components to Syntax

# FAQ

## What is the `Heading` component for?

Heading enforces a consistent style & accessibility best practices for headings.

## `Typography` is quite different from what we do in `Cambly-Frontend`, why is that?

* Sizes now go from small (100) to big (800) - previously the smallest size was `body300` which didn't make a lot of sense
* We now see `Typography` is the root component for all text to be rendered on Cambly. It allows for custom components such as `Heading` and has more flexibility built-in.

# Screenshots

## Typography

typography-overview

## Heading

heading-overview

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,10 @@
 {
+  "css.lint.validProperties": ["composes"],
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true,
+    "source.fixAll.stylelint": true
+  },
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true,
   "typescript.tsdk": "node_modules/typescript/lib"
 }
--- a/packages/syntax-core/src/Typography/Typography.stories.tsx
+++ b/packages/syntax-core/src/Typography/Typography.stories.tsx
@@ -1,0 +1,143 @@
+import { StoryObj, Meta } from "@storybook/react";
+import Typography from "./Typography";
+
+export default {
+  title: "Typography",
+  component: Typography,
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/p7LKna9JMU0JEkcKamzs53/%F0%9F%93%90-Syntax?node-id=4003-11352&t=t9iBh7rbKNhWCMrt-0",
+    },
+  },
+  argTypes: {
+    align: {
+      options: ["start", "center", "end", "forceLeft", "forceRight"],
+      control: { type: "radio" },
+      defaultValue: "start",
+    },
+    as: {
+      options: ["div", "h1", "h2", "h3", "h4", "h5", "h6"],
+      control: { type: "radio" },
+    },
+    color: {
+      options: [
+        "destructive-primary",
+        "gray700",
+        "gray900",
+        "primary",
+        "white",
+      ],
+      control: { type: "radio" },
+    },
+    inline: {
+      control: "boolean",
+    },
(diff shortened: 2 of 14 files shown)


.vscode/settings.json
```diff
@@ -1,3 +1,10 @@
 {
+  "css.lint.validProperties": ["composes"],
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true,
```
christianvuerings, line 4: Updated these to autofix ESLint / Stylelint and Prettier on save

packages/syntax-core/src/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,143 @@
...
+  component: Typography,
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/p7LKna9JMU0JEkcKamzs53/%F0%9F%93%90-Syntax?node-id=4003-11352&t=t9iBh7rbKNhWCMrt-0",
+    },
+  },
+  argTypes: {
+    align: {
+      options: ["start", "center", "end", "forceLeft", "forceRight"],
```
jliotta, line 15: What do `forceLeft` and `forceRight` do? What will they do when a language is rtl?

packages/syntax-core/src/Typography/Typography.tsx
```diff
@@ -0,0 +1,127 @@
...
+      return colorStyles.gray900Color;
+  }
+}
+
+/**
+ * Typography is a component that renders text.
+ */
+const Typography = ({
+  align = "start",
+  as = "div",
```
jliotta, line 29: Is `as` the standard way in design systems to designate which element to use? Just curious since I hadn't seen it before

somethiiing: i think it is, I've seen this around

somethiiing:
did you run changeset on this? it said no changeset found and was just double checking

packages/syntax-core/src/Typography/Typography.stories.tsx
christianvuerings: @jliotta see the description in on the `align` prop:

>   Aligns the text to the left, right, or center of the container.
>   * `start` and `end` will align the text to the left or right of the container depending on the locale.
>   * `forceLeft` and `forceRight` will align the text to the left or right of the container regardless of the locale (should be used sparingly).

We basically want to encourage folks to use `start` and `end` and only use `forceLeft` / `forceRight` in edge cases

christianvuerings:
> did you run changeset on this? it said no changeset found and was just double checking

@somethiiing added - thank you

packages/syntax-core/src/Typography/Typography.tsx
christianvuerings: @jliotta depends on the design system:

* MUI uses `component`: https://mui.com/material-ui/api/box/
* NextUI uses `as`: https://nextui.org/docs/layout/container#default
* ChakraUI uses: `as`: https://chakra-ui.com/docs/components/box/usage#as-prop

So there is no consensus yet but we are also not re-inventing the wheel

somethiiing (approved):

ajmooo:
@christianvuerings 700 and 800 should be using font-weight:860 (Heavy) instead of bold. I understand this only affects Apple people lol

ajmooo (approved):

PR #62, opened 2023-03-24
Checkbox w/ label component

nathan-ong:
Screenshot 2023-03-27 at 1 37 09 PM
Screenshot 2023-03-27 at 1 37 16 PM
Screenshot 2023-03-27 at 1 37 27 PM
Screenshot 2023-03-27 at 1 37 35 PM


packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.module.css
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.label {
+  display: flex;
+  flex-direction: row;
+  align-items:center;
+}
+
+
+/** TODO Change with `Typography` component when available */
```
christianvuerings, line 87: @nathan-ong that component is now available in Syntax

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.module.css
```diff
@@ -0,0 +1,105 @@
...
+.labelText {
+  color: inherit;
+  font-family: -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", Roboto,
+    Helvetica, Arial, sans-serif;
+  font-weight: 500;
+}
+
+.errorText {
+  color: #D32A4B;
```
christianvuerings, line 96: @nathan-ong should we use a color from our design tokens? Looks like this one is `destructive-700`

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.module.css
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.labelTextSmall {
+  font-size: 12px;
```
christianvuerings, line 100: @nathan-ong could we use `size` on `Typography` instead?

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { expect } from "vitest";
+
+describe("checkbox", () => {
+  it("renders successfully", () => {
+    const { baseElement } = render(
+      <LabeledCheckbox
+        checked
+        label="test"
+        onClick={() => {
+          /* empty */
```
christianvuerings, line 12: @nathan-ong would be great if we could write an actual test for this

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.tsx
```diff
@@ -0,0 +1,123 @@
...
+import classNames from "classnames";
+import styles from "./LabeledCheckbox.module.css";
+
+const iconSize = {
+  sm: styles.smallIconContainer,
+  md: styles.mediumIconContainer,
+};
+
+const textVariant = {
+  // Replace with `Typography` once it lands in `syntax-core`
```
christianvuerings, line 11: @nathan-ong just landed, so we can now use it

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.tsx
```diff
@@ -0,0 +1,123 @@
...
+          <div className={checkboxStyling}>
+            <div className={classNames(styles.iconContainer, iconSize[size])}>
+              <svg
+                className={styles.icon}
+                xmlns="http://www.w3.org/2000/svg"
+                fill="none"
+              >
+                <path
+                  fill="#fff"
+                  d="M5.313 10c-.14 0-.278-.028-.417-.083a1.187 1.187 0 0 1-.396-.271L.854 6a1.083 1.083 0 0 1-.323-.823c.007-.326.122-.6.344-.823.222-.222.493-.333.813-.333.319 0 .59.11.812.333l2.854 2.854L12.208.354c.223-.222.49-.333.802-.333.313 0 .58.11.803.333.222.222.333.49.333.802 0 .313-.111.58-.334.802L6.125 9.646a1.187 1.187 0 0 1-.396.27 1.115 1.115 0 0 1-.416.084Z"
```
christianvuerings, line 100: @nathan-ong we should be able to further optimize this with https://jakearchibald.github.io/svgomg/

package.json
```diff
@@ -25,7 +25,7 @@
     "stylelint": "15.2.0",
     "stylelint-config-css-modules": "4.2.0",
     "stylelint-config-standard": "30.0.1",
-    "turbo": "latest",
+    "turbo": "^1.8.6",
```
christianvuerings, line 28: @nathan-ong curious why we had to make this change. Did you run `pnpm install` instead of `npm install`?

nathan-ong: Ahh, yeah my linter was bugging out. I think I may have resolved it, so I'll revert this

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.tsx
nathan-ong: I believe I did pass it in to svgo, but I'll give it another try

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { ReactElement, useState } from "react";
+import classNames from "classnames";
+import styles from "./LabeledCheckbox.module.css";
+import Typography from "../Typography/Typography";
```
JFChoe, line 4: Should we use absolute imports everywhere instead?

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.tsx
```diff
@@ -0,0 +1,104 @@
...
+              <path
+                fill="#fff"
+                d="m9 16.2-3.5-3.5a.9839.9839 0 0 0-1.4 0c-.39.39-.39 1.01 0 1.4l4.19 4.19c.39.39 1.02.39 1.41 0L20.3 7.7c.39-.39.39-1.01 0-1.4a.9839.9839 0 0 0-1.4 0L9 16.2z"
+              ></path>
+            </svg>
+          )}
+        </div>
+      </div>
+      <Typography
+        size={typographySize[size]}
```
JFChoe, line 95: I'm getting a typescript error from this `No index signature with a parameter of type 'string' was found on type '{ sm: number; md: number; }'.` . But it works fine. Wonder if it's a bug w/ typescript? https://stackoverflow.com/questions/56568423/typescript-no-index-signature-with-a-parameter-of-type-string-was-found-on-ty

nathan-ong: @JFChoe I think this might be the fix for that: https://stackoverflow.com/questions/66993264/what-does-the-as-const-mean-in-typescript-and-what-is-its-use-case

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.module.css
```diff
@@ -0,0 +1,75 @@
...
+  border-radius: 6px;
+}
+
+.md {
+  height: 24px;
+  width: 24px;
+  border-radius: 8px;
+}
+
+.iconContainer {
```
christianvuerings, line 65: @nathan-ong do we use this class anywhere? If not, can we remove it?

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.module.css
```diff
@@ -0,0 +1,75 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.label {
```
christianvuerings, line 71: @nathan-ong same here, can we remove the class?

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+    },
+    size: {
+      options: ["sm", "md"],
+      control: { type: "radio" },
+    },
+    disabled: {
+      control: "boolean",
+    },
+    onClick: { action: "clicked" },
+    onKeyDown: { action: "hit space" },
```
christianvuerings, line 26: @nathan-ong Can we remove this since it's no longer a prop on LabeledCheckbox?

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+      checked={isChecked}
+      onChange={handleChange}
+      label="checkbox label"
+    />
+  );
+};
+
+export const Interactive: StoryObj<typeof LabeledCheckbox> = {
+  render: () => (
+    <>
```
christianvuerings, line 52: @nathan-ong Can we remove the fragment? I don't think we need it anymore

packages/syntax-core/src/LabeledCheckbox/LabeledCheckbox.tsx
christianvuerings: @JFChoe Not in Syntax - we still have to set up absolute imports

christianvuerings (approved):
LGTM after comments get addressed - thanks for adding this component @nathan-ong

christianvuerings:
@nathan-ong Just merged - now this will automatically kick off a deploy

PR #73, opened 2023-03-29
Select dropdown

nathan-ong:
@christianvuerings this still needs tests, but I wanted to get this out to see if you had any thoughts on how this is being implemented
Screenshot 2023-03-29 at 3 23 43 PM
Screenshot 2023-03-29 at 3 23 54 PM
Screenshot 2023-03-29 at 3 24 06 PM
Screenshot 2023-03-29 at 3 24 17 PM


packages/syntax-core/src/SelectList/SelectList.module.css
```diff
@@ -0,0 +1,78 @@
...
+  display: flex;
+  position: relative;
+}
+
+.selectBox {
+  width: 100%;
+  padding: 0 12px;
+  display: flex;
+  align-items: center;
+  border: 1px solid #767676;
```
christianvuerings, line 25: @nathan-ong could we use the `--color-base-gray-700` for this? https://cambly-syntax.vercel.app/?path=/docs/colors--docs

packages/syntax-core/src/SelectList/SelectList.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+import userEvent from "@testing-library/user-event";
+
+const SelectDropdown = ({
+  numberOfOptions,
+  onChange,
+}: {
+  numberOfOptions: number;
+  onChange: () => void;
+}) => {
+  const options = [];
```
christianvuerings, line 14: @nathan-ong a more functional way to write this:

```js
(new Array(numberOfOptions)).fill(0).map((item, index) => ({
  value: `value-${index + 1}`,
  label: `label-${index + 1}`
}))
```

Also differentiates between a `label` and `values`

packages/syntax-core/src/SelectList/SelectList.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+    options.push({ value: String(i), label: String(i) });
+  }
+  return (
+    <SelectList
+      selectedValue=""
+      onChange={onChange}
+      placeholderText="placeholder"
+      label="Select"
+    >
+      <>
```
christianvuerings, line 25: @nathan-ong can we remove this extra fragment?

packages/syntax-core/src/SelectList/SelectList.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+        {options.map((o) => (
+          <SelectList.Option key={o.value} value={o.value} label={o.value} />
```
christianvuerings, line 27: `label`, should we `label={o.label}` instead of `label={o.value}`. Also, you can destructure the arguments:

```jsx
{options.map(({value, label}) => (
  <SelectList.Option key={value} value={value} label={label} />
))}

File: packages/syntax-core/src/SelectList/SelectList.test.tsx [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+    expect(screen.getAllByRole("option").length).toBe(numberOfOptions + 1);
+  });
+  it("calls onchange on selecting option", async () => {
+    const handleChange = vi.fn();
+    render(<SelectDropdown onChange={handleChange} numberOfOptions={3} />);
+    await userEvent.selectOptions(screen.getByTestId("test-select-id"), "1");
+    await userEvent.selectOptions(screen.getByTestId("test-select-id"), "2");
+    expect(handleChange).toHaveBeenCalledTimes(2);
+  });
+  it("updates selections when selecting multiple options  ", async () => {
```
christianvuerings, line 67: Could we remove the extra spaces after `options`? `options"`

packages/syntax-core/src/SelectList/SelectList.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    // Expect one additional option because of the placeholder option
+    expect(screen.getAllByRole("option").length).toBe(numberOfOptions + 1);
+  });
+  it("calls onchange on selecting option", async () => {
+    const handleChange = vi.fn();
+    render(<SelectDropdown onChange={handleChange} numberOfOptions={3} />);
+    await userEvent.selectOptions(screen.getByTestId("test-select-id"), "1");
+    await userEvent.selectOptions(screen.getByTestId("test-select-id"), "2");
+    expect(handleChange).toHaveBeenCalledTimes(2);
```
christianvuerings, line 65: @nathan-ong would be good to also check what `handleChange` has been called with:

```js
expect(handleChange).toHaveBeenCalledWith(...
```

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -0,0 +1,136 @@
...
+  disabled?: boolean;
+  /**
+   * true if there is an error with the selection
+   */
+  error?: boolean;
+  /**
+   * Text shown below select box if there is an input error.
+   * Should only have a value if error = true
+   */
+  errorText?: string;
```
christianvuerings, line 41: @nathan-ong could we remove `error` and only use `errorText`? If `errorText` is set, then we know there is an error

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * Text shown below select box
+   */
+  helperText?: string;
+  /**
+   * Text shown above select box
+   */
+  label?: string;
```
christianvuerings, line 49: @nathan-ong should `label` be required so we make the select accessible?

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * The callback to be called when an option is selected
+   */
+  onChange: React.ChangeEventHandler<HTMLSelectElement>;
+  /**
+   * Text showing in select box if no option has been chosen
+   *
+   */
+  placeholderText: string;
```
christianvuerings, line 58: @nathan-ong should this be optional?

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * Value of the currently selected option
+   */
+  selectedValue: string;
```
christianvuerings, line 62: @nathan-ong should this be optional?

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -0,0 +1,136 @@
...
+          size={100}
+          color={error ? "destructive-primary" : "gray700"}
+        >
+          {label}
+        </Typography>
+      </label>
+      <div className={styles.selectWrapper}>
+        <select
+          id={id}
+          data-testid="test-select-id"
```
christianvuerings, line 92: @nathan-ong can we prefix the testid with `syntax`? `data-testid="syntax-select"`

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -0,0 +1,136 @@
...
+        </select>
+        <div className={styles.arrowIcon}>
+          <svg
+            focusable="false"
+            aria-hidden="true"
+            viewBox="0 0 24 24"
+            width={iconSize[size]}
+          >
+            <path
+              fill={error ? "#d32a4b" : "#353535"}
```
christianvuerings, line 116: @nathan-ong could we import these colors from `syntax-design-tokens`?

e.g.

```js
import { ColorBaseDestructive700 } from "@cambly/syntax-design-tokens/dist/js/index.js";
```

packages/syntax-core/src/SelectList/SelectOption.tsx
```diff
@@ -0,0 +1,21 @@
...
+  value,
+  label,
+  disabled = false,
+}: {
+  value: string;
+  label: string;
+  disabled?: boolean;
+}): ReactElement => (
+  <option
+    data-testid={`select-option-${value}`}
```
christianvuerings, line 13: @nathan-ong same here, could we prefix the test id with `syntax`?

```jsx
data-testid={`syntax-select-value-${value}`}
```

packages/syntax-core/src/SelectList/SelectList.test.tsx
nathan-ong: Ahh very cool. Is there a need to fill with `0` since we're just using it to map over?

nathan-ong: I was looking into this a little more and came across this way of doing it 
```
Array.from({length:numberOfOptions},(item, index)=> ({value: index + 1, label: index + 1}))
```
Slightly more space optimized because we're not creating an additional array by mapping over the first array?

packages/syntax-core/src/SelectList/SelectList.tsx
nathan-ong: Yeah, I thought about this. Can it be safely assumed that there will always be an error text upon error?

nathan-ong: I would think so, but I'm curious if, for example if we had a form with multiple dropdowns and all fields were required, if we would show the error text individually on each dropdown, or if there would just be an error message on the parent that contained these dropdowns, and we would just make these dropdowns red without an error for each of them?

packages/syntax-core/src/SelectList/SelectList.tsx
nathan-ong: Yeah, good call, I don't think there's reason to not have a label

packages/syntax-core/src/SelectList/SelectList.tsx
nathan-ong: Or, are you saying if there's no placeholder, we default to showing option 1?

packages/syntax-core/src/SelectList/SelectList.tsx
nathan-ong: I just figured if the parent is tracking the selectedValue, then we're always going to need to pass a `selectedValue` prop to the `SelectList`, right?

packages/syntax-core/src/SelectList/SelectList.tsx
nathan-ong: ooh yes, I didn't think to do it this way!

nathan-ong:
@christianvuerings I think this is ready for another look. Looks like importing the design tokens is causing some issues both with the tests and the linter

packages/syntax-core/src/SelectList/SelectList.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+  },
+  argTypes: {
+    size: {
+      options: ["sm", "md", "lg"],
+      control: { type: "radio" },
+    },
+    disabled: {
+      control: "boolean",
+    },
+    error: {
```
christianvuerings, line 24: @nathan-ong could we remove `error`, I don't think that exists anymore

packages/syntax-core/src/SelectList/SelectList.test.tsx
```diff
@@ -0,0 +1,99 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
christianvuerings, line 1: @nathan-ong could we inline these disables per line?

That way we
* can count them 
* avoid accidentally disabling one

packages/syntax-core/src/SelectList/SelectList.module.css
```diff
@@ -0,0 +1,80 @@
...
+  flex-direction: column;
+  gap: 8px;
+}
+
+.opacityOverlay {
+  opacity: 0.3;
+}
+
+.outerTextContainer {
+  padding-inline-start: 4px;
```
christianvuerings, line 12: @nathan-ong should we also add padding to the end? That way we'll be using symmetrical padding

```jsx
padding-inline-end: 4px;
padding-inline-start: 4px;
```

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -0,0 +1,135 @@
...
+          disabled={disabled}
+          className={classNames(styles.selectBox, styles[size], {
+            [styles.unselected]: !selectedValue && !errorText,
+            [styles.selected]: selectedValue && !errorText,
+            [styles.selectError]: errorText,
+          })}
+          onChange={onChange}
+        >
+          {placeholderText && (
+            <option disabled aria-selected value="">
```
christianvuerings, line 101: @nathan-ong should we remove `aria-selected`?

christianvuerings (approved):

packages/syntax-core/src/SelectList/SelectList.test.tsx
nathan-ong: ah, ok that's good to know the reasoning behind that

PR #111, opened 2023-04-10
Button: update border styles for secondary + destructive secondary.

somethiiing:
# Updating Border Styles for secondary + destructive secondary to match new figma designs:

## After SS:
image
image

## Before SS:
image
image

### DIFF
--- a/packages/syntax-core/src/Button/Button.tsx
+++ b/packages/syntax-core/src/Button/Button.tsx
@@ -107,9 +107,17 @@ const Button = ({
 }): ReactElement => {
   const contextProps = useContext(ButtonGroupContext);
 
-  const size = contextProps?.size ?? sizeProp;
-  const disabled = contextProps?.disabled ?? disabledProp;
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/Button/Button.tsx
```diff
@@ -107,9 +107,9 @@ const Button = ({
 }): ReactElement => {
   const contextProps = useContext(ButtonGroupContext);
 
-  const size = contextProps?.size || sizeProp;
-  const disabled = contextProps?.disabled || disabledProp;
-  const fullWidth = contextProps?.fullWidth || fullWidthProp;
+  const size = contextProps?.size ?? sizeProp;
```
christianvuerings, line 110: @somethiiing I don't think we can use `??` in this case since the button should not be disabled when `disabled` is `false`

somethiiing: fixed by adding comments + disabling the warnings for using `||`

christianvuerings (approved):

PR #116, opened 2023-04-12
Select list error state

nathan-ong:
Gray label, w/ red placeholder and error text

Screenshot 2023-04-12 at 2 39 57 PM

### DIFF
--- a/packages/syntax-core/src/SelectList/SelectList.tsx
+++ b/packages/syntax-core/src/SelectList/SelectList.tsx
@@ -93,9 +93,12 @@ const SelectList = ({
             [styles.selectError]: errorText,
           })}
           onChange={onChange}
+          value={
+            placeholderText && !selectedValue ? placeholderText : selectedValue
+          }
         >
           {placeholderText && (
-            <option disabled value="" selected>
+            <option disabled value={placeholderText}>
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -98,7 +95,7 @@ const SelectList = ({
           onChange={onChange}
         >
           {placeholderText && (
-            <option disabled value="">
+            <option disabled value="" selected>
```
christianvuerings, line 98: @nathan-ong If we have a `selectedValue` and a placeholder, we want the `selectedValue` to be selected, not the `option`.

So I would change it to something like this:
```tsx
<select
  //
  value={
    placeholderText && !selectedValue ? placeholderText : selectedValue
  }
>
  {placeholderText && (
    <option disabled value={placeholderText}>
      {placeholderText}
    </option>
  )}
```

That way:
* If we have `placeholderText` but no `selectedValue`, the placeholder still gets selected
* If we have `placeholderText` and `selectedValue`, the correct option gets selected

christianvuerings (changes requested):

christianvuerings (approved):

PR #124, opened 2023-04-18
Hide select label

nathan-ong:


packages/syntax-core/src/SelectList/SelectList.module.css
```diff
@@ -18,6 +18,10 @@
   position: relative;
 }
 
+.visuallyHidden {
+  display: none;
```
christianvuerings, line 22: @nathan-ong `display: none` doesn't just visually hide it, it also hides it from screen readers.

Instead I recommend you use `<Box display="visuallyHidden" />`

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -20,6 +20,7 @@ const SelectList = ({
   errorText,
   helperText,
   label,
+  hideLabel = false,
```
christianvuerings, line 23: @nathan-ong instead of a boolean prop, how do you feel about:

```tsx
labelDisplay: "visible" | "hidden"
```

nathan-ong: ahh, ok, yeah that makes sense to me!

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -44,9 +46,14 @@ const SelectList = ({
    */
   helperText?: string;
   /**
-   * Text shown above select box
+   * Text shown above select box or for screen readers
    */
   label: string;
+  /**
+   *  "visible": label displays above select dropdown
+   *  "hidden": label hidden from display
```
christianvuerings, line 54: @nathan-ong could you add a `@defaultValue`?

```jsx
@defaultValue `visible`
```

christianvuerings (approved):
Approved on the condition that we address the defaultValue comment

PR #125, opened 2023-04-20
Divider: allow key to be set as a prop

christianvuerings:
# Changes

Allow `key` to be set as a prop on `<Divider />`

# Why

Any React component should accept `key` as a prop, otherwise we always have to wrap it if we use it in a list


nathan-ong (approved):
Thanks for adding!

PR #130, opened 2023-04-21
Box: Add flex-shrink

jliotta:
For certain languages, the text inside of `Button` was stacking. Adding flex-shrink 0 fixes the issue

Before:
image

After:
Screenshot 2023-04-21 at 3 14 19 PM

### DIFF
--- a/packages/syntax-core/src/Box/Box.tsx
+++ b/packages/syntax-core/src/Box/Box.tsx
@@ -124,17 +124,16 @@ export default function Box(props: {
    */
   display?: Display;
   /**
+   *  If the size of all flex items is larger than the flex container, items shrink to fit according to flex-shrink
+   *
+   */
+  flexShrink?: number;
+  /**
    * By default, flex items will all try to fit onto one line. But if you specify `flexWrap="wrap"`, the flex items will wrap onto multiple lines.
    *
    * @defaultValue `nowrap`
    */
   flexWrap?: "wrap" | "nowrap";
-  /**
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/Button/Button.module.css
```diff
@@ -98,6 +98,7 @@
 }
 
 .textContainer {
+  flex-shrink: 0;
```
christianvuerings, line 101: @jliotta instead of setting it on `Button`, can we wrap that button with a `<Box />` which has `flex-shrink: 0`?

Asking because in certain cases, we do want the text to wrap:
image

Otherwise we are going to get into a situation where we see horizontal scrollbars:

image

packages/syntax-core/src/Box/Box.tsx
```diff
@@ -129,6 +129,12 @@ export default function Box(props: {
    * @defaultValue `nowrap`
    */
   flexWrap?: "wrap" | "nowrap";
+  /**
+   *  If the size of all flex items is larger than the flex container, items shrink to fit according to flex-shrink
+   *
+   * @defaultValue `nowrap`
```
christianvuerings, line 135: @jliotta `nowrap` doesn't make sense as a default value, should we just remove it?

christianvuerings (approved):
Approved but would still be good to see the comments addressed

packages/syntax-core/src/Box/Box.tsx
jliotta: oh yeah that was from copying the flexWrap template

PR #135, opened 2023-05-01
Typography: Add lineClamp

jliotta:
This PR adds a `lineClamp` property to `Typography` to allow us to truncate the text at a specified number of lines

Screenshot 2023-05-01 at 4 23 42 PM

### DIFF
--- a/packages/syntax-core/src/Typography/Typography.tsx
+++ b/packages/syntax-core/src/Typography/Typography.tsx
@@ -30,6 +30,7 @@ const Typography = ({
   children,
   color = "gray900",
   inline = false,
+  lineClamp,
   size = 200,
   tooltip,
   transform = "none",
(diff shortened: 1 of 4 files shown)


packages/syntax-core/src/Typography/Typography.tsx
```diff
@@ -66,6 +67,12 @@ const Typography = ({
    * @defaultValue false
    */
   inline?: boolean;
+  /**
+   * The number of lines we should truncate the text at
+   *
+   * @defaultValue null
```
christianvuerings, line 73: @jliotta we can remove the `@defaultValue` if it is `null` or `undefined`

christianvuerings (approved):

packages/syntax-core/src/Typography/Typography.tsx
```diff
@@ -30,6 +30,7 @@ const Typography = ({
   children,
   color = "gray900",
   inline = false,
+  lineClamp = undefined,
```
christianvuerings, line 33: @jliotta this will be the default already, so no need to set it again.

So just `lineClamp`

PR #136, opened 2023-05-02
Box: pass through 'role' attribute

christianvuerings:
# Changes

Pass through `role` attribute on `Box`

# Why

Helps with `Box` adoption and ensuring we have all accessibility attributes available


zprong:
🙌 That's just pulling main and running `npm install` correct? If so, then I'm good!

zprong:
^ Scratch that - but Joe helped me with the upgrade :) Thanks again for adding this!

PR #146, opened 2023-05-22
Internal: enable a few more TypeScript ESLint rules

christianvuerings:
# Changes

* Enables a few more TypeScript ESLint rules to enforce consistent type imports / explicit boundaries and only use functions after they have been defined


JFChoe (approved):
lgtm

PR #150, opened 2023-05-24
Badge: Add component

jliotta:
Add a badge component

[Figma designs](https://www.figma.com/file/p7LKna9JMU0JEkcKamzs53/%F0%9F%93%90-Syntax?type=design&node-id=3954-11404&t=FoeFM80KltQbwuYV-0)


jliotta:
@christianvuerings I have a couple questions on my implementation: 

1) Should I be adding new color helping functions for `badgeBackgroundColor` and `badgeForegroundColor` like this? 
2) I had some issues with adding the background colors to Box so I used a span. I'd prefer to use Box. In that case should I modify `backgroundColor` and `foregroundColor` with the new cases?

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,44 @@
...
+   */
+  text: string;
+  /**
+   * The color of the badge
+   *
+   * @defaultValue "primary"
+   */
+  color?: (typeof BadgeColor)[number];
+}): JSX.Element => (
+  <span
```
christianvuerings, line 27: @jliotta Could we make this a `div` instead which will make it `display: block`? If folks do want to show it next to something else, it'll encourage them to use `flex`

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,44 @@
...
+    className={classNames(
+      badgeBackgroundColor(color),
+      badgeForegroundColor(color),
+    )}
+    style={{
+      display: "inline",
+      borderRadius: "100px",
```
christianvuerings, line 34: @jliotta Can we move these styles to a CSS modules file for `Badge`?

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,44 @@
...
+      padding: "4px 9px",
```
christianvuerings, line 35: @jliotta Could you check with @ajmooo to see if we can make this `4px 8px` instead?

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,44 @@
...
+    }}
+  >
+    <Typography color="inherit" inline>
```
christianvuerings, line 38: @jliotta Just as above, let's remove the `inline`

packages/syntax-core/src/colors/badgeBackgroundColor.ts
```diff
@@ -0,0 +1,23 @@
+import { type BadgeColor } from "../constants";
+import styles from "./colors.module.css";
+
+export default function badgeBackgroundColor(
```
christianvuerings, line 4: Not a must but feel free to inline this function in `Badge.tsx` since we only use it within that file. In general it makes sense to abstract it out if it's used in 2+ places

christianvuerings (commented):
Looks great - just a few comments

christianvuerings:
> I had some issues with adding the background colors to Box so I used a span. I'd prefer to use Box. In that case should I modify backgroundColor and foregroundColor with the new cases?

@jliotta I would indeed try to use a `<Box />` when possible. What issues did you run in with `<Box />`?

jliotta:
> > I had some issues with adding the background colors to Box so I used a span. I'd prefer to use Box. In that case should I modify backgroundColor and foregroundColor with the new cases?
> 
> @jliotta I would indeed try to use a `<Box />` when possible. What issues did you run in with `<Box />`?

Applying the background and foreground color since I can't use `className` like in IconButton and Button

packages/syntax-core/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,5 @@
+.badge {
+  border-radius: 100px;
+  padding: 4px 8px;
+  display: inline-block;
```
jliotta, line 4: @christianvuerings If I don't do some sort of `inline` display then the badge takes up the entire width available instead of just the width of the text plus padding

christianvuerings: @jliotta yes, should be fine - thank you

packages/syntax-core/src/constants.ts
```diff
@@ -14,4 +14,14 @@ export const Color = [
   "white",
   "inherit",
 ] as const;
+
+export const BadgeColor = [
```
christianvuerings, line 18: @jliotta these `constants` should be for things that are truly global, not just for a specific component like `Badge`.

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,40 @@
...
+  color?: (typeof BadgeColor)[number];
+}): JSX.Element => (
+  <div
+    className={classNames(
+      styles.badge,
+      badgeBackgroundColor(color),
+      badgeForegroundColor(color),
+    )}
+  >
+    <Typography color="inherit">{text}</Typography>
```
christianvuerings, line 36: @jliotta should the typography not be size 100 and bold?

christianvuerings (commented):
@jliotta please take a look at https://github.com/jliotta/syntax/pull/1 which is a PR to your badge branch. It converts everything to use `Box` and removes the unnecessary files

christianvuerings (approved):
@jliotta looks good - just be sure to add `Badge` to `packages/syntax-core/src/index.tsx` so it's exported

PR #162, opened 2023-06-08
SelectList: render helperText when errorText is an empty string

christianvuerings:
# Changes

* SelectList: render helperText when errorText is an empty string
* Add tests

### DIFF
--- a/packages/syntax-core/src/SelectList/SelectList.tsx
+++ b/packages/syntax-core/src/SelectList/SelectList.tsx
@@ -143,7 +143,7 @@ const SelectList = ({
           size={100}
           color={errorText ? "destructive-primary" : "gray700"}
         >
-          {errorText ?? helperText}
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -143,7 +143,7 @@ const SelectList = ({
           size={100}
           color={errorText ? "destructive-primary" : "gray700"}
         >
-          {errorText ?? helperText}
+          {errorText ? errorText : helperText}
```
christianvuerings, line 146: We use this syntax instead of `errorText || helperText` to avoid triggering https://typescript-eslint.io/rules/prefer-nullish-coalescing/

zprong (approved):
Looks great!! Thanks so much for updating this Christian :)

PR #167, opened 2023-06-15
Modal: Add Modal + Components to Syntax

somethiiing:


packages/syntax-core/src/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+  it("should render footer buttons successfully", async () => {
+    render(
+      <Modal
+        header="title"
+        onDismiss={() => {
+          /* empty */
+        }}
+        footer={<button>confirm</button>}
+      >
+        <p>text</p>
```
christianvuerings, line 77: @somethiiing use `data-testid` instead

packages/syntax-core/src/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+          /* empty */
+        }}
+      >
+        <p>text</p>
+      </Modal>,
+    );
+    expect(baseElement).toHaveStyle({ overflow: "hidden" });
+  });
+
+  it("should render FocusTrap succesfully", async () => {
```
christianvuerings, line 130: @somethiiing could we add a test to test `FocusTrap`? Seems like it's not working at the moment

packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+  );
+}
+
+// Note: Only sm + lg size currently. design thinks there should only be two sizes.
+// If there IS a md size at some point, we should use the "size" const.
+const ModalSizes = ["sm", "lg"] as const;
+
+type ModalType = {
+  /**
+   * The children inside for the inside of a Modal.
```
christianvuerings, line 27: @somethiiing Change to `Modal's main content`

packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+  image?: JSX.Element;
+  /**
+   * The footer for the bottom area of the Modal.
+   * Typically used for rendering buttons.
+   */
+  footer?: JSX.Element;
+  /**
+   * The size of the card
+   *
+   * `sm`: 400px
```
christianvuerings, line 53: @somethiiing Could we make these bullet points? Just like we do in other places in Syntax

https://github.com/Cambly/syntax/blob/7c1f90584c4adb7e771cfa4899b85876d245ec4a/packages/syntax-core/src/IconButton/IconButton.tsx#L28-L30

packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+   *
+   * @defaultValue sm
+   */
+  size?: (typeof ModalSizes)[number];
+  /**
+   * The z-index for the modal.
+   * Typically used if there are other things on the page with higher z-index and you need this overlayed on top.
+   * If two(2) buttons, put primary button first.
+   *
+   * @defaultValue 1
```
christianvuerings, line 64: @somethiiing Change the default to `0`, just like we do in `Layer`

packages/syntax-core/src/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+      <Modal
+        header="title"
+        onDismiss={() => {
+          /* empty */
+        }}
+      >
+        <p>text</p>
+      </Modal>,
+    );
+    const header = await screen.findAllByText("title");
```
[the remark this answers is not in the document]
somethiiing: this one doesn't seem like it can have a `data-testid`, because the text is passed into `heading` component. unless we want to add it to the component? (which i dont think we want to do)

packages/syntax-core/src/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,140 @@
+import { useState } from "react";
```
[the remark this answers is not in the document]
somethiiing: all of these are fixed except cancel/confirm im not sure if i can make it close the element because of how the structure of the jsx is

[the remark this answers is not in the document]
somethiiing: the onClick of the button needs to be passed in from the outside, but the function to close is defined inside, so i dont think its doable(?) without lots of restructuring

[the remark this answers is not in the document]
somethiiing: image

somethiiing:
handled most comments, missing focus trap behavior fix + functionality test. + 1 data test id question

packages/syntax-core/src/Modal/FocusTrap.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { useEffect, useRef, type ReactElement } from "react";
```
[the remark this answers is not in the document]
somethiiing: it works on other places, just not on storybook because of it loading in an iframe, will write a test for functionality tho

packages/syntax-core/src/Modal/Modal.stories.tsx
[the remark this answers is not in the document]
christianvuerings: @somethiiing Can we write it as follows:

```tsx
export const Default: StoryObj<typeof Modal> = {
  render: function DefaultExample({ ...args }): JSX.Element {
    const [isOpen, setIsOpen] = useState(false);

    return (
      <>
        <Button
          onClick={() => {
            setIsOpen(true);
          }}
          text={"Click here to open Modal"}
        />
        {isOpen && (
          <ModalTemplate
            {...args}
            onDismiss={() => setIsOpen(false)}
            header={args.header || "One Button Modal Text"}
            footer={
              <Button
                text="Confirm"
                onClick={() => {
                  action("confirm");
                  setIsOpen(false);
                }}
              />
            }
          />
        )}
      </>
    );
  },
};
```

it's fine to have a bit more repetition in these examples

packages/syntax-core/src/Modal/Modal.test.tsx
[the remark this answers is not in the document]
christianvuerings: @somethiiing makes sense, instead can we use `getByRole` - it also enforces that we use the correct heading level.

```tsx
expect(screen.getByRole("heading", { level: 1 })).toHaveTextContent(
  "title",
);
```

somethiiing:
handled remaining comments

packages/syntax-core/src/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,262 @@
...
+import { action } from "@storybook/addon-actions";
```
christianvuerings, line 2: @somethiiing should we increase the zIndex in the examples to e.g. `1` so we don't see the show code buttons?
image

image

packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -0,0 +1,164 @@
...
+              display="flex"
+              direction="column"
+              minWidth={240}
+              maxWidth={sizeWidth[size]}
+              width="100%"
+              dangerouslySetInlineStyle={{ __style: { overflow: "hidden" } }}
+            >
+              <Box position="relative">
+                <button
+                  aria-label={accessibilityCloseLabel}
```
christianvuerings, line 126: @somethiiing can we fix the focus styles on this button? We have this issue because we don't use `IconButton` in this case

packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * The size of the card
+   *
+   * * `sm`: 400px
+   * * `lg`: 600px
+   *
+   *
+   * @defaultValue sm
+   */
+
```
christianvuerings, line 73: @somethiiing remove extra line break here

packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * Test id for the modal
+   */
+  "data-testid"?: string;
+};
+
+/**
+ * Modal is a dialog that appears on top of the main content and locks user interaction within the modal.
+ */
+
```
christianvuerings, line 91: @somethiiing remove extra linebreak here

christianvuerings (approved):
@somethiiing Looks great - thanks for making the changes! Please address the comments before merging

ajmooo:
Sorry I've been slow to see these requests. I have some changes

- padding should be 36px, currently it's set to 40px
- modal body text is set to size100 and should be size200
- I'm not 100% sure we want to allow a modal with no buttons? Can talk about in person on Monday.

ajmooo (changes requested):
Added changes to a comment - not sure if it's more convenient for you to see them there or here… 

- padding should be 36px, currently it's set to 40px
- modal body text is set to size100 and should be size200
- I'm not 100% sure we want to allow a modal with no buttons? Can talk about in person on Monday.

ajmooo (approved):
thank you

PR #168, opened 2023-06-15
Add generate component script to Syntax

somethiiing:
This script helps consistency with generating components + should hopefully speed up syntax dev process.

No changelog because no file changes.

Run `pnpm generate ComponentName` to generate all the files.

### DIFF
--- a/generate-component/Template.test.tsx
+++ b/generate-component/Template.test.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { screen, render } from "@testing-library/react";
 import Template from "./Template";
 
+// NOTE: Update this to be the lowercased version!
 describe("Template", () => {
   it("it should render successfully", () => {
     // Update tests here:
--- a/generate-component/index.ts
+++ b/generate-component/index.ts
@@ -1,6 +1,5 @@
 import fsSync, { promises as fs } from "fs";
 import { parseArgs } from "node:util";
-import { component, story, test } from "./component_templates";
 
 function writeFileErrorHandler(err) {
   if (err) throw err;
@@ -84,10 +83,11 @@ async function generateComponent() {
     );
 
     // insert new component into 'packages/syntax-core/src/index.tsx file
-    updateIndexFile(name);
+    // updateIndexFile(name);
 
     console.log(
-      "Note: Formatting of files might change because of our lint rules when you save.",
+      `TODO: Update the describe block of your newly generated .test.tsx to be lowercased! e.g: describe("template", ....)\n`,
+      "Note: Formatting of files might change because of our lint rules when you save.\n",
       "Note 2: Make sure the casing of your Component names are 'StartCased' ex: 'ButtonGroup' or 'Card' ",
     );
   } catch (error) {


generate-component/index.js
```diff
@@ -0,0 +1,77 @@
+const fs = require("fs");
```
christianvuerings, line 1: @somethiiing can we use `import` here and `.ts`? You can use tsx to run the file:

```
npx tsx generate-component
```

generate-component/index.js
```diff
@@ -0,0 +1,77 @@
...
+const [name] = process.argv.slice(2);
+if (!name) throw new Error("You must include a component name.");
+
+const uppercasedName =
+  name.charAt(0).toUpperCase() + name.slice(1).toLowerCase();
+
+const dir = `./packages/syntax-core/src/${uppercasedName}/`;
+
+// throw an error if the file already exists
+if (fs.existsSync(dir))
```
christianvuerings, line 14: @somethiiing can we convert the `fs.*` to the async / promise versions?

generate-component/index.js
```diff
@@ -0,0 +1,77 @@
...
+    fs.writeFile(
+      `./packages/syntax-core/src/index.tsx`,
+      fileContent,
+      writeFileErrorHandler,
+    );
+  },
+);
+
+console.log(
+  "Note: Formatting of files might change because of our lint rules when you save.",
```
christianvuerings, line 76: @somethiiing as mentioned above, would be great if we can split those files out so we already apply prettier / ESLint to them

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "private": true,
   "scripts": {
+    "start": "turbo run dev --no-cache --continue",
```
christianvuerings, line 4: @somethiiing can you see if `"start": "dev"` or `"start": "pnpm run dev"` works? We want to make sure that these stay in sync

christianvuerings (commented):
@somethiiing Really like the idea - do know that these generators are usually tricky to maintain & ensure that they follow the latest coding standards

README.md
```diff
@@ -11,9 +11,16 @@ nvm install
 npm install -g pnpm
 pnpm install
 # Run the dev server
-pnpm dev
+pnpm start
 ```
 
+## Generate a new React Component
+
+```
```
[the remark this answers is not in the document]
somethiiing: oops

generate-component/index.js
somethiiing: in this particular case, i didn't really want ESlint rules to apply to them. (in my opinion at least)

in a component instead of:

```
<Box>
// I leave comment here
</Box>
```

I don't really want it to be converted to:

`<Box />`

I wanted to make it clear where the devs would make changes. But I'm open to a discussion

generate-component/index.js
somethiiing: will look into this

generate-component/index.js
```diff
@@ -0,0 +1,77 @@
+const fs = require("fs");
+const { component, story, test, barrel } = require("./component_templates.tsx");
+
+// grab component name from terminal argument
+const [name] = process.argv.slice(2);
+if (!name) throw new Error("You must include a component name.");
+
+const uppercasedName =
+  name.charAt(0).toUpperCase() + name.slice(1).toLowerCase();
```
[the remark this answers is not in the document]
somethiiing: I was trying to make it easier for someone to be able to type a "lowercased" component name and have it be properly cased. But I guess if it's a Two worded component name (like ButtonGroup) it would break.

I'll just revert these changes and add a note telling people to just properly name the component. and use whatever is passed in

generate-component/index.js
```diff
@@ -0,0 +1,77 @@
+const fs = require("fs");
+const { component, story, test, barrel } = require("./component_templates.tsx");
+
+// grab component name from terminal argument
+const [name] = process.argv.slice(2);
```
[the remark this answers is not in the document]
somethiiing: done

jliotta (approved):
awesome!

generate-component/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+// insert new component into 'packages/syntax-core/src/index.tsx file
+fs.readFile(
+  "./packages/syntax-core/src/index.tsx",
+  "utf8",
+  function (err, data) {
+    if (err) throw err;
+
+    // grab all components and combine them with new component
+    const currentComponents = data.match(/(?<=import )(.*?)(?= from)/g);
```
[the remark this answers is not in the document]
somethiiing: if this is okay, i'm going to leave it for now. learning how to AST will be a task to tackle another day

generate-component/Template.test.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import { screen, render } from "@testing-library/react";
+import Template from "./Template";
+
+describe("Template", () => {
```
somethiiing, line 5: The only problem with the "read file" from file approach is that I don't think I have a good way of forcing the describe block's `Template` to be lowercased.

I think leaving it like this is fine and adding a comment to handle it is worth the benefits of having the ts linted file.

generate-component/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  const name = positionals[0];
+  if (!name) throw new Error("You must include a component name.");
+
+  const dir = `./packages/syntax-core/src/${name}/`;
+
+  // Throw an error if the file already exists
+  if (fsSync.existsSync(dir))
+    throw new Error("A component with that name already exists.");
+
+  try {
```
somethiiing, line 55: I wasn't sure if each `read + writeFile` should be it's own `try catch` @christianvuerings do you have opinions on this? ideally we want it to fail if any of these fail,  so I think this is fine

PR #180, opened 2023-06-26
SelectList: pass through onClick

christianvuerings:
@zprong noticed an issue in the `SelectListDate.tsx` component with the #174 PR. In that PR, we rely on the event being propagated since it looks like a `select` but is actually acting as a datepicker:


zprong (approved):
Thanks for updating this!!

zprong:
@christianvuerings I forget.. did we ever find out why the dropdown's textbox didn't show up in the `SelectListDate` component?

PR #182, opened 2023-06-28
SelectList: Adjust font sizes according to size + update font weight + make label required

christianvuerings:
# Changes

* Adjust font sizes according to size
* Update font weight 
* Make label required

### DIFF
--- a/.changeset/warm-dots-add.md
+++ b/.changeset/warm-dots-add.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": major
+---
+
+SelectList: make label required
--- a/packages/syntax-core/src/SelectList/SelectList.module.css
+++ b/packages/syntax-core/src/SelectList/SelectList.module.css
@@ -19,17 +19,19 @@
 }
 
 .selectBox {
-  width: 100%;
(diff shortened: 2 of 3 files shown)


ajmooo:
@christianvuerings 
1. when I select with my keyboard in arc (chrome based) i see a blue outline. Is that just a browser thing? 
2. Can you confirm the placeholder text is medium? I know browsers render type differently from Figma and I had trouble inspecting the specific text inside the component. 
3. This looks awesome otherwise!

Screenshot 2023-06-29 at 10 13 56 AM

ajmooo (approved):
I have some clarifying questions in a comment, approved otherwise!

christianvuerings:
> @christianvuerings
> 
> 1. when I select with my keyboard in arc (chrome based) i see a blue outline. Is that just a browser thing?
> 2. Can you confirm the placeholder text is medium? I know browsers render type differently from Figma and I had trouble inspecting the specific text inside the component.
> 3. This looks awesome otherwise!

1. Yes - removed it though since we already have the black focus style
2. Yes - we used the same font-weight as a button

PR #184, opened 2023-07-04
Button / IconButton: make onClick optional

christianvuerings:
# Changes

Button / IconButton: make onClick optional

# Why?

See `Button` test - it allows for a `Button` with `type="submit"` to submit a form

### DIFF
--- a/.changeset/fluffy-masks-invite.md
+++ b/.changeset/fluffy-masks-invite.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+Button / IconButton: make onClick optional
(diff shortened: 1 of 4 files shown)


benjroy (approved):
LGTM


Aside and only tangentially related at best: this makes me wonder if we have/should have a component like `ButtonLink`/`LinkButton` (or have `Button` take an `href` that sets up an `onClick()` to navigate to the `href`

Unsure if that should be a desing-system component or if it is just frontend-application level component.  since we have both `express` and `next` apps going on, that might suggest it should be app component.

We had an `href` prop on the design-system Button at nerdwallet.  It was both super-handy and also a pain in the rear change.

Ok, done with my distracting curiosity here.

🚢 :shipit: 🚢

christianvuerings:
@benjroy `ButtonLink`/`LinkButton` => Yes, that is actually something @somethiiing is currently working on. The tricky bit is to support both Express react-router-dom and the Next.js router without unnecessarily increasing our bundle size

PR #187, opened 2023-07-06
TextField: add component

christianvuerings:
# Changes

Add the `TextField` component which currently lives in `Cambly-Frontend`

# Screenshot

image

### DIFF
--- a/packages/syntax-core/src/TextField/TextField.tsx
+++ b/packages/syntax-core/src/TextField/TextField.tsx
@@ -1,0 +1,137 @@
+import React, {
+  type ReactElement,
+  type HTMLInputTypeAttribute,
+  useId,
+} from "react";
+import classNames from "classnames";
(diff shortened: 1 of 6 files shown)


packages/syntax-core/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,137 @@
...
+  type HTMLInputTypeAttribute,
+  useId,
+} from "react";
+import classNames from "classnames";
+import styles from "./TextField.module.css";
+import Box from "../Box/Box";
+import Typography from "../Typography/Typography";
+
+/**
+ * [TextField](https://cambly-syntax.vercel.app/?path=/docs/components-textfield--docs) is a component that allows users to enter text.
```
christianvuerings, line 12: #188 is a follow up where we add these links to other components

somethiiing (approved):

PR #188, opened 2023-07-06
Docs: add individual links to each component

christianvuerings:
# Changes

Add a link to the storybook for each component

# Why?

Make it way easier to navigate to the docs for a specific component

image

### DIFF
--- a/.changeset/tame-knives-fold.md
+++ b/.changeset/tame-knives-fold.md
@@ -1,0 +1,6 @@
+---
+"@cambly/syntax-floating-components": patch
+"@cambly/syntax-core": patch
+---
+
+Add individual links to components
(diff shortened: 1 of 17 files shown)


.changeset/tame-knives-fold.md
```diff
@@ -0,0 +1,6 @@
+---
+"@cambly/syntax-floating-components": patch
+"@cambly/syntax-core": patch
+---
+
+Add individual links to components
```
benjroy, line 6: Aside, newb question:
👀 ok this file has me curious.  what system is using this and why is it titled `tame-knives-fold`? 👀

benjroy (approved):
100% _YES_

Did you just add these changes, or do you have some local scripts that you use to apply changes across multiple files?

.changeset/tame-knives-fold.md
christianvuerings: @benjroy See https://github.com/changesets/changesets - that name gets auto generated.

https://github.com/changesets/changesets/blob/main/docs/common-questions.md#each-changeset-is-its-own-file

> We use random human readable names by default for these files to avoid collisions when generating them, but there's no harm that will come from renaming them.

PR #192, opened 2023-07-07
force pnpm usage

benjroy:
This PR uses the `engines` field in `package.json` and the `engine-strict` config in `.npmrc` to force the usage of `pnpm` in this repo.   It will throw an error if `npm` is used.

Please note:  this rewrote the lockfile when `pnpm install` was run again

Note: `packageManager` field in `package.json` is for [corepack](https://nodejs.org/api/corepack.html) usage, which is experimental node feature that has to be opted into.  That could be useful for us to explore longterm.

### DIFF
--- a/.changeset/curly-mirrors-dress.md
+++ b/.changeset/curly-mirrors-dress.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": patch
+---
+
+Change root monorepo package.json to force usage of pnpm at specific ^version
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -11,7 +11,7 @@ runs:
       name: Install pnpm
       id: pnpm-install
       with:
-        version: 7
+        version: 8
         run_install: false
     - name: Get pnpm store directory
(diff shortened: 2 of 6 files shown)


benjroy:
I can't test out the storybook preview because I don't have access somewhere:

Clicking the link takes me to vercel auth page:
image

Logging in with github takes me to an account-not-found page:

image


Does a switch need to be flipped somewhere on our end to allow my user access?

christianvuerings (approved):

benjroy:
Just noticed something weird the the repo DX in here:

I had auto-merge enabled for this PR.  Got an approval from you @christianvuerings , but it couldn't auto-merge because there was a conflict in the lockfile.

I then resolved the conflict in that lockfile and pushed up.  It immediately automerged when conflict was resolved based on your previous approval, ⚠️ _but it merged BEFORE the ci test run was complete_ ⚠️ 

That felt _dangerous_ , and has me wondering if this is something we can tweak in the auto-merge setup or if we need to tick the box in github repo settings to dismiss previous approvals when new commits are added.

🤔 🤔 🤔

PR #199, opened 2023-07-11
IconButton: make accessibilityLabel required

christianvuerings:
# Changes

**BREAKING CHANGE!**
Make accessibilityLabel required on `IconButton`

# Why?

Enforces every IconButton to have a correct accessibility label

Avoids issues like this:
image

### DIFF
--- a/.changeset/heavy-moons-report.md
+++ b/.changeset/heavy-moons-report.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": major
+---
+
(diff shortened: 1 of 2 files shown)


benjroy (approved):
For a moment I was wondering if we could automatically apply the button text as accessibilityLabel, then realized this is an IconButton with no text.

And we can't automatically apply the `tooltip` text, since that is not guaranteed to be supplied

PR #200, opened 2023-07-11
Feat: Typography transform="capitalize"

benjroy:
# Changes

Make `<Typography />` accept `transform="capitalize"`

# Why?

Working on implementing a design for a new `<LessonSchedulingRep />`:
https://www.figma.com/file/Y9IVAFSurWopQ7d6O2ucdx/Lesson-Rep?type=design&node-id=1-2877
image

The badges capitalize the words inside.
This particular text comes back as lowercase strings.
This allows them to display as designed without having to split apart the string and transform it in JS.

After:
image
image

Before:
image
image

### DIFF
--- a/.changeset/small-jeans-relate.md
+++ b/.changeset/small-jeans-relate.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+Typography: add "capitalize" as potential value for 'transform' prop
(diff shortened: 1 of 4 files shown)


benjroy:
I'm going to close this because I wanted to use it inside of a `Badge`, but right now Badge only takes child text as a `string` and I don't want to expand that interface just for this.

PR #205, opened 2023-07-13
Add CheckboxGroup component

zprong:
image

image


packages/syntax-core/src/CheckboxGroup/CheckboxGroup.stories.tsx
```diff
@@ -0,0 +1,65 @@
+import { type StoryObj, type Meta } from "@storybook/react";
```
christianvuerings, line 1: @zprong can you make sure to run `pnpm run changeset` locally to add a changeset? This should be a minor change for `@cambly/syntax-core`

packages/syntax-core/src/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const onChange = (
+    { target: { checked } }: React.ChangeEvent<HTMLInputElement>,
+    id: string,
+  ) => {
+    const updatedSelections = { ...selections, [id]: checked };
+    onFormChange(updatedSelections);
+  };
+
+  return (
+    <Box display="flex" direction="column" gap={3}>
```
christianvuerings, line 52: @zprong ideally folks would be able to use `CheckboxGroup` as follows:

```tsx
<CheckboxGroup>
  <Checkbox label="soccer">Soccer</Checkbox>
  <Checkbox label="baseball">Baseball</Checkbox>
  <Checkbox label="basketball">Basketball</Checkbox>
</CheckboxGroup>
```

That makes it more composable (everything is still a React element) and avoid use having to pass in big objects to components.

We can make `CheckboxGroup` dumb and not support other props so it would just be:
```tsx
<Box display="flex" direction="column" gap={3}>
  {children}
</Box>
```

If you do want it to be smarter, I would recommend using React context to pass the state around from `CheckboxGroup` to `Checkbox`.

A good example to look at:
* https://react-spectrum.adobe.com/react-spectrum/CheckboxGroup.html
* https://github.com/adobe/react-spectrum/blob/main/packages/%40react-spectrum/checkbox/src/CheckboxGroup.tsx

PR #207, opened 2023-07-20
LinkButton: add rel prop for best practices

somethiiing:
Missing rel prop for best practices.

Depending on the discussion on what happens re: `camblystrap` overriding css for `LinkButton` as well, might need a `!important`, but waiting on that to merge this.

### DIFF
--- a/packages/syntax-core/src/LinkButton/LinkButton.module.css
+++ b/packages/syntax-core/src/LinkButton/LinkButton.module.css
@@ -1,4 +1,5 @@
-.linkButton {
+.linkButton:focus,
+.linkButton:hover {
   text-decoration: none;
 }
 
--- a/packages/syntax-core/src/LinkButton/LinkButton.module.css
+++ b/packages/syntax-core/src/LinkButton/LinkButton.module.css
@@ -1,3 +1,7 @@
+.linkButton {
+  text-decoration: none;
+}
+
 .linkButton:focus,
 .linkButton:hover {
   text-decoration: none;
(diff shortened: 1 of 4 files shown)


packages/syntax-core/src/LinkButton/LinkButton.tsx
```diff
@@ -45,6 +46,19 @@ export default function LinkButton({
    *
    */
   target?: "_blank" | "_self" | "_parent" | "_top";
+  /**
+   * The target attribute specifies where to open the linked document.
+   *
+   */
+  rel?:
```
christianvuerings, line 53: @somethiiing this should be changed to `rel?: HTMLAttributes<HTMLAnchorElement>["rel"]` which will resolve to `rel?: string` since `rel` can take multiple values:
https://www.elegantthemes.com/blog/wordpress/rel-noopener-noreferrer-nofollow

packages/syntax-core/src/LinkButton/LinkButton.module.css
```diff
@@ -1,4 +1,5 @@
-.linkButton {
```
christianvuerings, line 1: @somethiiing we also need `.linkButton` here since it's the default styling https://syntax-git-linkbutton-add-rel-prop-for-best-practices-cambly.vercel.app/?path=/docs/components-linkbutton--docs

image

christianvuerings (approved):
Approving on the condition that the comments will be addressed

PR #213, opened 2023-07-31
Feat: `<Badge icon={IconComponent} />`

benjroy:
## What

This PR adds an `icon` prop to `<Badge />`.
- ex: `<Badge icon={RestartIcon} text="Restart" />`

It allows for the rendering of an IconComponent before badge text string

@somethiiing your comment [here](https://github.com/Cambly/Cambly-Frontend/pull/6084#discussion_r1279202650) helped me decide how to implement this 🙏 🙏 

## Why

The design I am implementing calls for an Icon displayed in a Badge:  https://www.figma.com/file/Y9IVAFSurWopQ7d6O2ucdx/Scheduling-Rep?node-id=159%3A6571&mode=dev

Screenshots:

Badge in figma:
image

syntax storybook:
image

### DIFF
--- a/packages/syntax-core/src/Badge/Badge.module.css
+++ b/packages/syntax-core/src/Badge/Badge.module.css
@@ -1,0 +1,13 @@
+.icon {
+  /* stylelint-disable-next-line declaration-no-important --  Need to use !important to override MUI icon sizes  */
+  height: 16px !important;
+  /* stylelint-disable-next-line declaration-no-important --  Need to use !important to override MUI icon sizes  */
+  width: 16px !important;
+
+  /**
+   * The badge size is 22px total, text is size 14px, there is 4px top + 4px bottom padding
+   * Adjust the margins on the (optional) icon so that is displays the right size but does not affect overall height
+   */
+  margin-top: -1px;
+  margin-bottom: -1px;
+}
--- a/packages/syntax-core/src/Badge/Badge.tsx
+++ b/packages/syntax-core/src/Badge/Badge.tsx
@@ -1,6 +1,5 @@
 import Typography from "../Typography/Typography";
 import Box from "../Box/Box";
-import { type ReactElement } from "react";
 import styles from "./Badge.module.css";
 
 const BadgeColor = [
--- a/packages/syntax-core/src/Badge/Badge.tsx
+++ b/packages/syntax-core/src/Badge/Badge.tsx
@@ -1,5 +1,7 @@
 import Typography from "../Typography/Typography";
 import Box from "../Box/Box";
+import { type ReactElement } from "react";
+import styles from "./Badge.module.css";
 
 const BadgeColor = [
   "gray200",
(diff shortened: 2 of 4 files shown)


packages/syntax-core/src/Badge/Badge.module.css
```diff
@@ -0,0 +1,13 @@
...
+  height: 16px !important;
+  /* stylelint-disable-next-line declaration-no-important --  Need to use !important to override MUI icon sizes  */
+  width: 16px !important;
+
+  /**
+   * The badge size is 22px total, text is size 14px, there is 4px top + 4px bottom padding
+   * Adjust the margins on the (optional) icon so that is displays the right size but does not affect overall height
+   */
+  margin-top: -1px;
+  margin-bottom: -1px;
```
benjroy, line 12: 👀 The design calls for badge of always 22px height.  4px top + bottom padding, 14px line-height, 12px font-size, with a 16px w/h Icon size.

The margins here are adjusted by -1px top -1px bottom in order to keep the 16px icon from affecting the overall height of the badge.

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -28,13 +30,18 @@ const textColorForBackgroundColor = (
...
 }: {
+  /**
+   * The icon to be displayed. Please use a [Material Icon](https://material.io/resources/icons/)
+   */
+  icon?: React.ComponentType<{ className: string }>;
   /**
    * The text to display inside the badge
    */
-  text: string;
+  text: string | ReactElement;
```
benjroy, line 44:
```suggestion
  text: string;
```

Nope, not a ReactElement.  started off that direction, then went with `icon` prop after hearing opinions from @nathan-ong @somethiiing @jliotta

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -48,13 +55,19 @@ const Badge = ({
     paddingY={1}
     rounding="full"
     backgroundColor={color}
+    dangerouslySetInlineStyle={{ __style: { lineHeight: "14px" } }}
```
benjroy, line 58: 👀 👀 The design calls for badge of always 22px height. 4px top + bottom padding, 14px line-height, 12px font-size, with a 16px w/h Icon size.

Typography does not set line-height for any given size (I think it should!), so I am hard-coding the 14px line-height here.

Otherwise it is calculated to 15px, which makes the overall height of the Badge 23px instead of 22px

somethiiing (approved):

packages/syntax-core/src/Badge/Badge.module.css
christianvuerings: @benjroy did you confirm with @ajmooo whether this is actually what we want? Ideally the height of the badge would be divisible by 4 - so either `24px` or `20px` high.

Also, the `-1` is a bit of a hack - let's see if we can work with design to avoid it

benjroy: Thank you!  Have not yet confirmed that this is what we actually want.  The `-1` is definitely a hack, and your review is just what I need to give a poke to update the `Badge` in figma!

PR #220, opened 2023-08-11
Update Avatar to include a 1px white border

zprong:
Discussion is in [this slack](https://cambly.slack.com/archives/C033ZPY5M46/p1691699159828119)

Background set to black for this screenshot for clarity
image

As AJ noted, the 1px should be an _internal stroke_ so the size shouldn't increase as a result of adding the white border. Inspected that they're the same as they were both - as an example, a medium-sized avatar is still 40px

image

### DIFF
--- a/packages/syntax-core/src/Avatar/Avatar.module.css
+++ b/packages/syntax-core/src/Avatar/Avatar.module.css
@@ -1,27 +1,31 @@
+:root {
+  --syntax-avatar-border-width: 1px;
+}
+
 .avatar {
   background-position: center;
   background-size: cover;
   background-repeat: no-repeat;
   border-radius: 50%;
-  border: 1px solid #fff;
+  border: var(--syntax-avatar-border-width) solid #fff;
 }
 
 .sm {
-  width: 22px;
-  height: 22px;
+  width: calc(24px - var(--syntax-avatar-border-width) * 2);
+  height: calc(24px - var(--syntax-avatar-border-width) * 2);
 }
 
 .md {
-  width: 38px;
-  height: 38px;
+  width: calc(40px - var(--syntax-avatar-border-width) * 2);
+  height: calc(40px - var(--syntax-avatar-border-width) * 2);
 }
 
 .lg {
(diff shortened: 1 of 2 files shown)


jliotta (approved):

packages/syntax-core/src/Avatar/Avatar.module.css
```diff
@@ -3,24 +3,25 @@
   background-size: cover;
   background-repeat: no-repeat;
   border-radius: 50%;
+  border: 1px solid #fff;
 }
 
 .sm {
-  width: 24px;
-  height: 24px;
+  width: 22px;
```
christianvuerings, line 10: @zprong It's not very clear why we use `22px` instead of `24px` - I know it's the calculation of `24px` (image width) - the border on each side (`2 * 1px`) resulting in `22px`.

Instead, we could use CSS variables to make this more explicit:

```css
:root {
  --syntax-avatar-border-width: 1px;
}

.avatar {
  background-position: center;
  background-size: cover;
  background-repeat: no-repeat;
  border-radius: 50%;
  border: var(--syntax-avatar-border-width) solid #fff;
}

.sm {
  width: calc(24px - var(--syntax-avatar-border-width) * 2);
  height: calc(24px - var(--syntax-avatar-border-width) * 2);
}

.md {
  width: calc(40px - var(--syntax-avatar-border-width) * 2);
  height: calc(40px - var(--syntax-avatar-border-width) * 2);
}

.lg {
  width: calc(72px - var(--syntax-avatar-border-width) * 2);
  height: calc(72px - var(--syntax-avatar-border-width) * 2);
}

.xl {
  width: calc(128px - var(--syntax-avatar-border-width) * 2);
  height: calc(128px - var(--syntax-avatar-border-width) * 2);
}
```

This way, if we ever change the border width to e.g. `2px`, we don't have to do the calculations again ourselves.

zprong: Brilliant! I'll incorporate this, thanks Christian!

PR #221, opened 2023-08-11
Typography: add interactive weight

jliotta:
Some of AJ's designs require a specific 500 font-weight for interactive text. Currently I've named that variant `interactive` but not sure if we want another name that is less tied to its use and more tied to a descriptor like `regular` and `semiBold`

### DIFF
--- a/.changeset/modern-ants-reply.md
+++ b/.changeset/modern-ants-reply.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+typography: add interactive font-weight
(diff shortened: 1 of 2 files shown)


christianvuerings:
@ajmooo do we want `500 font-weight for interactive text` to be used by every engineer? Or is that option only for internal interactive components like `Button` or `Link`?

ajmooo (approved):
This is great. My initial thinking is that this text style would only be used for syntax components, but i think it makes sense to have it live with all other typography like you're suggesting. Just need to have specific rules around how/when to use so designers don't start using it for reasons outside indicating interactivity.

somethiiing (approved):

PR #231, opened 2023-08-21
[Popover] Create component + Add Elevation

somethiiing:
- Add elevation
- Add Popover component + stories + tests

image


emmanuelgamboa09:
@christianvuerings if you have a chance can you look this over since im going to need it for the navbar and I haven't really reviewed stuff on syntax yet so not to sure on what we generally look out for here

packages/syntax-floating-components/src/Popover/Popover.module.css
```diff
@@ -0,0 +1,7 @@
+.popoverContent {
+  box-sizing: border-box;
+  background-color: var(--color-base-white);
+  padding: 28px;
+  border-radius: 24px;
+  width: 352px;
```
emmanuelgamboa09, line 6: Ideally I would like to pass in the width I want. But one thing to note this will be annoying on devices smaller than 352px. If we are forced to set a width i think it makes sense to set this to maxWidth of 352px. Just so if it needs to shrink down it can. But again ideally would like to pass in my own widths and maxWidths lol.

emmanuelgamboa09: maybe atleast make lg md sm size popovers or something lol.

somethiiing: oh we can get rid of sizes, we discussed that with aj yesterday. ill get rid of it now

somethiiing: itll take in a width that you can pass in

emmanuelgamboa09: maxWidth and width field please

emmanuelgamboa09 (approved):
lgtm. May want other eyes on it but seems good to me.

PR #234, opened 2023-08-29
Chip: Add Component

ChessyHsu:
Add chip component + stories + tests

https://github.com/Cambly/syntax/assets/17971529/402265ab-2f89-4370-9c2d-e8af2f9c803c


packages/syntax-core/src/Chip/Chip.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const handleClick = vi.fn();
+    render(
+      <Chip
+        data-testid="button-test-id"
+        onChange={handleClick}
+        text="text on chip"
+      />,
+    );
+    expect(screen.getByTestId("button-test-id")).toBeInTheDocument();
+  });
```
christianvuerings, line 59: @ChessyHsu could you also add a check for the `aria-pressed` value?

```tsx
  it("sets the correct pressed state", async () => {
    function ChipInteractive() {
      const [selected, setSelected] = useState(false);

      return (
        <Chip
          data-testid="button-test-id"
          onChange={() => setSelected(!selected)}
          selected={selected}
          text="text on chip"
        />
      );
    }
    render(<ChipInteractive />);
    const chip = await screen.findByTestId("button-test-id");
    expect(chip).toHaveAttribute("aria-pressed", "false");
    await userEvent.click(chip);
    expect(chip).toHaveAttribute("aria-pressed", "true");
  });
```

christianvuerings (commented):
@ChessyHsu Thanks for adding this component! Please add a changeset by running `pnpm changeset` (should be a minor release)

christianvuerings (approved):
@ChessyHsu looks great! Thanks for making all the updates

PR #242, opened 2023-09-07
Chip: Update chip to have cursor on hover

somethiiing:
Chip is a clickable element and should have a cursor on hover. Adding `cursor: pointer` should fix this issue.

https://www.loom.com/share/14724ecbb04f499999bed4ed8b46374b


joemanley201 (approved):

ChessyHsu:
LGTM!!

PR #243, opened 2023-09-07
LinkButton: Update box-sizing

somethiiing:
LinkButton previously did not have `box-sizing: border-box` which was causing it to overflow (and not be centered) in certain places of the UI. Adding that should solve it.

Example: (note: this isn't live, just a demo)

BEFORE: 
image

AFTER: 
image

### DIFF
--- a/packages/syntax-core/src/LinkButton/LinkButton.module.css
+++ b/packages/syntax-core/src/LinkButton/LinkButton.module.css
@@ -1,6 +1,6 @@
 .linkButton {
-  text-decoration: none;
   box-sizing: border-box;
+  text-decoration: none;
 }
 
 .linkButton:focus,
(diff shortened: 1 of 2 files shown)


packages/syntax-core/src/LinkButton/LinkButton.module.css
```diff
@@ -1,5 +1,6 @@
 .linkButton {
   text-decoration: none;
+  box-sizing: border-box;
```
christianvuerings, line 3: @somethiiing small nit - ideally we alphabetize these properties when possible.

christianvuerings (approved):

packages/syntax-core/src/LinkButton/LinkButton.module.css
somethiiing: got it, fixed

PR #246, opened 2023-09-09
TapArea: add component

christianvuerings:
# Changes

Adds the `TapArea` component to Syntax

# Why?

We see more and more of a need of an accessible component which is clickable / tapable (to support mobile). Just grepping for `role="button"` in Cambly-Frontend shows 33 instances:
image

# Notes

* The `compress` feature can happen in a follow up - I just added a base component that folks can start using today

# Screenshot

image

### DIFF
--- a/packages/syntax-core/src/Box/Box.module.css
+++ b/packages/syntax-core/src/Box/Box.module.css
@@ -106,27 +106,6 @@
   align-self: auto;
 }
 
-/* Border Radius */
-.roundingsm {
-  border-radius: 8px;
-}
-
-.roundingmd {
-  border-radius: 12px;
-}
-
-.roundinglg {
-  border-radius: 16px;
-}
-
-.roundingxl {
-  border-radius: 32px;
-}
-
-.roundingfull {
-  border-radius: 999px;
-}
-
 /* Display (base) */
 .block {
   display: block;
--- a/packages/syntax-core/src/TapArea/TapArea.module.css
+++ b/packages/syntax-core/src/TapArea/TapArea.module.css
@@ -1,0 +1,38 @@
+.tapArea {
+  box-sizing: border-box;
+}
+
+.fullWidth {
+  width: 100%;
+}
+
+.disabled {
+  filter: opacity(50%);
+  background-image: none;
+  transform: none;
+  cursor: auto;
+}
+
+.enabled {
+  cursor: pointer;
+}
+
+.enabled:hover {
+  background-image: linear-gradient(
+    to bottom,
+    rgba(0, 0, 0, 0.1) 0%,
+    rgba(0, 0, 0, 0.1) 100%
+  );
+  transition-duration: 0.2s;
+  cursor: pointer;
+}
+
+.enabled:focus-visible {
+  background-image: linear-gradient(
+    to bottom,
+    rgba(0, 0, 0, 0.2) 0%,
+    rgba(0, 0, 0, 0.2) 100%
+  );
+  box-shadow: 0 0 0 4px #000;
+  outline: solid 2px #fff;
+}
--- a/packages/syntax-core/src/TapArea/TapArea.stories.tsx
+++ b/packages/syntax-core/src/TapArea/TapArea.stories.tsx
@@ -1,0 +1,84 @@
+import { type StoryObj, type Meta } from "@storybook/react";
+import TapArea from "./TapArea";
+import Avatar from "../Avatar/Avatar";
+import image from "../../../../apps/storybook/assets/images/jane.webp";
+import Box from "../Box/Box";
+import Typography from "../Typography/Typography";
+
+export default {
+  title: "Components/TapArea",
+  component: TapArea,
+  argTypes: {
+    disabled: {
+      control: "boolean",
+    },
+    onClick: { action: "clicked" },
+    rounding: {
+      options: ["none", "sm", "md", "lg", "xl", "full"],
+      control: { type: "select" },
+    },
+    tabIndex: {
+      control: { type: "number" },
+      options: [0, -1],
+    },
+  },
+  tags: ["autodocs"],
+} as Meta<typeof TapArea>;
+
+export const Default: StoryObj<typeof TapArea> = {
+  args: {
+    children: (
+      <Box display="flex" alignItems="center" gap={4} padding={2}>
+        <Avatar accessibilityLabel="Jane" src={image} size="md" />
+        <Typography size={500}>Jane Doe</Typography>
+      </Box>
+    ),
+  },
+};
+
+export const Disabled: StoryObj<typeof Default> = {
+  args: { ...Default.args, disabled: true },
+};
+
+const roundingLookup = {
+  sm: "8px",
+  md: "12px",
+  lg: "16px",
+  xl: "32px",
+  full: "999px",
+} as const;
+
+export const Rounding: StoryObj<typeof Box> = {
+  render: () => (
+    <>
+      <Box display="flex" direction="column" gap={4}>
+        <Typography>Hover to see the rounding effect</Typography>
+
+        <Box
+          display="flex"
+          gap={4}
+          flexWrap="wrap"
+          backgroundColor="gray100"
+          padding={4}
+        >
+          {(["sm", "md", "lg", "xl", "full"] as const).map((rounding) => (
+            <TapArea
+              key={rounding}
+              rounding={rounding}
+              fullWidth={false}
+              onClick={() => {
+                /* empty */
+              }}
+            >
+              <Box display="flex" alignItems="center" gap={2} padding={2}>
+                <Typography tooltip={`${roundingLookup[rounding]}`}>
+                  rounding=&quot;{rounding}&quot;
+                </Typography>
+              </Box>
+            </TapArea>
+          ))}
+        </Box>
+      </Box>
+    </>
+  ),
+};
--- a/packages/syntax-core/src/TapArea/TapArea.test.tsx
+++ b/packages/syntax-core/src/TapArea/TapArea.test.tsx
@@ -1,0 +1,170 @@
+import { screen, render } from "@testing-library/react";
+import TapArea from "./TapArea";
+import { expect, vi } from "vitest";
+import userEvent from "@testing-library/user-event";
+import { createRef } from "react";
+
+describe("tapArea", () => {
+  it("renders successfully", () => {
+    const { baseElement } = render(
+      <TapArea
+        onClick={() => {
+          /* empty */
+        }}
+      />,
+    );
+    expect(baseElement).toBeTruthy();
+  });
+
+  it("renders an role=button element", async () => {
+    render(
+      <TapArea
+        onClick={() => {
+          /* empty */
+        }}
+      />,
+    );
+    const tapArea = await screen.findAllByRole("button");
+    expect(tapArea).toHaveLength(1);
+  });
+
+  it("renders its children", async () => {
+    render(
+      <TapArea
+        onClick={() => {
+          /* empty */
+        }}
+      >
+        <div data-testid="tap-area-child-testid">Continue</div>
+      </TapArea>,
+    );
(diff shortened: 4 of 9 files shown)


packages/syntax-core/src/TapArea/TapArea.tsx
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * If `true`, the tap area will be disabled
+   *
+   * @defaultValue false
+   */
+  disabled?: boolean;
+  /**
+   * If `true`, the tap area will be full width
+   */
+  fullWidth?: boolean;
```
somethiiing, line 27: If TapArea doesn't have a size prop, is it worth considering to always make this full width? I haven't really thought this entirely through, but I imagine we could control this similar to how `Divider` is, which is the full size of the parent

packages/syntax-core/src/TapArea/TapArea.module.css
```diff
@@ -0,0 +1,59 @@
...
+  background-image: none;
+  transform: none;
+  cursor: auto;
+}
+
+.enabled {
+  cursor: pointer;
+}
+
+.enabled:hover {
```
somethiiing, line 20: Not entirely sure if this is true, but I think I saw a few of joey's designs and not all of them necessarily needed a hover state on all of the clickable cards. 

Maybe worth considering to have hover styles be a prop that can be toggled off (like disabled)

somethiiing (approved):
A couple of things that might be worth discussing, but looks good regardless

packages/syntax-core/src/TapArea/TapArea.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+      <TapArea
+        data-testid="tap-area-testid"
+        onTap={() => {
+          /* empty */
+        }}
+        accessibilityLabel="Continue to the next step"
+        fullWidth
+      />,
+    );
+    const tapArea = await screen.findAllByLabelText(
```
somethiiing, line 70: I noticed that some places use `findByTestId` and some use `findAllByLabelText`. Do you use `findAllByLabelText` for testing when you need to need something done externally for it?

and `findByTestId`, when you need to see the internals of the component?

ajmooo:
What's the color that's being used for hover state? Can that be the same color we're using for button hover, which I think is 10% #000?

packages/syntax-core/src/TapArea/TapArea.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+  args: { ...Default.args, disabled: true },
+};
+
+const roundingLookup = {
+  sm: "8px",
+  md: "12px",
+  lg: "16px",
+  xl: "32px",
+  full: "999px",
+} as const;
```
benjroy, line 49: Aside: outside of this PR, what are your thoughts on starting to consolidate these kind of common prop values that are starting to be duplicated into more and more component implementations?

i.e. thoughts on sharing styles and interfaces instead of duplicating them?

Off the top of my head I'm thinking rounding, borders, focus-rings, etc...

packages/syntax-core/src/TapArea/TapArea.tsx
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * The callback to be called when the tap area is clicked
+   */
+  onTap: React.MouseEventHandler<HTMLDivElement>;
```
benjroy, line 31: Why did you choose to call this `onTap`? 

The touch events are all named `onTouch...` https://developer.mozilla.org/en-US/docs/Web/API/Touch_events

is `onTap` a name that will make it easier to conceptualize for the implementations in ios/android?

At first I was thinking this might be the name that `react-native` chooses to use for this kind of handler, but nope, that's [onPress](https://reactnative.dev/docs/handling-touches)

So why `onTap`?

benjroy (commented):
Nice work, and thanks for taking this on.  I'll swing back for a full review, just wanted to ask a few questions to start.

I think the props-interface looks like a great set of restrictions to start out with.

On a side note, this seems to basically be an ariakit [Focusable](https://ariakit.org/components/focusable) with hover styling.  If we choose to adopt AriaKit in this repo, then if we like we can easily wire in that usage behind this props-interface :+1::+1:

packages/syntax-core/src/TapArea/TapArea.tsx
christianvuerings: @somethiiing in the case of `TapArea`, we probably don't want it to be always full width, otherwise, you would always need to set a container around it (probably a `Box`) when using it inline.

Example from the docs where we want to show them side-by-side:

image
image

packages/syntax-core/src/TapArea/TapArea.module.css
christianvuerings: @somethiiing We can make the default to have hover styles for now - if we do need it, we can always add a prop to disable them. However, in this case, I would lean to having more consistency across our UX / codebase

packages/syntax-core/src/TapArea/TapArea.test.tsx
christianvuerings: @somethiiing I used both in these tests just to make sure both worked (`data-testid` and label text).

christianvuerings:
> What's the color that's being used for hover state? Can that be the same color we're using for button hover, which I think is 10% #000?

@ajmooo that is correct, it's the same a button - we also apply a 0.2 second animation to it

packages/syntax-core/src/TapArea/TapArea.stories.tsx
christianvuerings: @benjroy yes, we can start to consolidate those as we see more duplication. I took a stab at `rounding` as an example, the other ones can happen in a follow up.

I am always a bit hesitant with over-abstracting these from the beginning since it does mean it means each component is a bit less isolated and requires you to start looking at more files.

packages/syntax-core/src/TapArea/TapArea.tsx
christianvuerings: @benjroy because of the name **Tap**Area & because this will work across mobile and desktop.

I agree with you though so I changed it back to `onClick` - if we need `onTouch*` events, we can add those on an as-needed basis.

christianvuerings:
> Nice work, and thanks for taking this on. I'll swing back for a full review, just wanted to ask a few questions to start.
> 
> I think the props-interface looks like a great set of restrictions to start out with.
> 
> On a side note, this seems to basically be an ariakit [Focusable](https://ariakit.org/components/focusable) with hover styling. If we choose to adopt AriaKit in this repo, then if we like we can easily wire in that usage behind this props-interface 👍👍

I tried out using `Focusable` instead by importing `Focusable` from `@ariakit/react` and these were the raw bundle sizes:

| Description | Bundle size | % increase | 
| --- | --- | --- |
| Before `TapArea` | 87K | / |
| With `TapArea` | 88K | 1.15% |
| With `TapArea` using `Focusable` | 104.6K | 20.2% |

So even for a relatively small component it would mean a 20.2% increase, that seems excessive. Also, to put this into context, our Next.js pages are around 200kB by themselves. So an extra 20kB does have an impact on load performance.

packages/syntax-core/src/Box/Box.module.css
```diff
@@ -106,27 +106,6 @@
   align-self: auto;
 }
 
-/* Border Radius */
```
christianvuerings, line 109: Moved to `rounding.module.css` to consolidate

packages/syntax-core/src/TapArea/TapArea.stories.tsx
benjroy: Excellent.  As an aside, I'll keep my eye out for other things to consolidate in future work.  The first thing that comes to mind (nothing to do with this PR) is `borderStyles` so we can eventually add a border prop interface to `Box`

packages/syntax-core/src/TapArea/TapArea.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+    expect(child).toHaveTextContent("Continue");
+  });
+
+  it("correctly applies fullWidth when set", async () => {
+    render(
+      <TapArea
+        data-testid="tap-area-testid"
+        onClick={() => {
+          /* empty */
+        }}
```
benjroy, line 51: aside, non-blocking, side-discussion:

Seeing all the empty `onClick` handlers in the test here makes we wonder if we shouldn't make `onClick` an optional prop.  `onClick` is always optional for the HTMLElements that implements it.

packages/syntax-core/src/TapArea/TapArea.tsx
```diff
@@ -0,0 +1,96 @@
...
+        aria-disabled={disabled}
+        aria-label={accessibilityLabel}
+        className={classNames(
+          styles.tapArea,
+          styles[`${disabled ? "disabled" : "enabled"}`],
+          fullWidth && styles.fullWidth,
+          rounding !== "none" && roundingStyles[`rounding${rounding}`],
+        )}
+        data-testid={dataTestId}
+        onClick={handleClick}
```
benjroy, line 83: This component needs to have `onKeydown` defined to run `onClick` handler on `Enter / Spacebar` keypress, in order to meet accessibility spec, doesn't it?

benjroy (changes requested):
Thanks for the thoughtful discussion responses!

I think that the component needs to handle `onKeydown` for accessibility reasons before this merges.  After that this is getting my ✅ ✅ 


I'f I'm wrong about that, lmk and I'll remove the 🚫 and add a ✅

benjroy:
This is a side-discussion whose resolution should not hold up this PR in any way:

> I tried out using Focusable instead by importing Focusable from @ariakit/react and these were the raw bundle sizes:

@christianvuerings yea, I think that it would be premature to ship this component built on top of `@ariakit/Focusable` at this moment since it would be the first introduction of that library anywhere in our app bundles, so the bundle cost will look unreasonably high like you have just pointed out.

I think it'll make more sense once our apps are already shipping bits of ariakit.  At that point, `Focusable` will already be in the app bundles (Focusable is a lower-level abstraction in ariakit, and thus used by almost all parts of the library that we'll be shipping), and the additional bundle-size cost should be minimal.

And we should be able to reap back the size of the ariakit dependencies in our app bundles when we ultimately get to remove the `material-ui` dependencies they'll replace.


Oh, one more thing, the comments and concerns I've been hearing re: bundle-sizes & syntax:  It finally sank in last week that the built js we ship from `dist` in our packages are all bundled into a single javascript file, so requiring one bit of syntax pulls the entire library into the app bundle.

This PR should make the library more tree-shakeable and allow us to start thinking about bundle-size puzzles in a more isolated way:  https://github.com/Cambly/syntax/pull/249/files

packages/syntax-core/src/TapArea/TapArea.tsx
```diff
@@ -0,0 +1,96 @@
...
+const TapArea = forwardRef<HTMLDivElement, TapAreaProps>(
+  (
+    {
+      children,
+      accessibilityLabel,
+      "data-testid": dataTestId,
+      disabled = false,
+      fullWidth = true,
+      onClick,
+      rounding = "none",
```
benjroy, line 64: might want to double-check with design it we prefer a little rounding as default, because changing default behavior in the future could take a little more mental energy once this is used throughout the apps.

packages/syntax-core/src/TapArea/TapArea.tsx
christianvuerings: @benjroy Yes, agreed - let's tackle that in a follow up as you mentioned

packages/syntax-core/src/TapArea/TapArea.tsx
christianvuerings: @benjroy yes - will follow up with @ajmooo everywhere else though the `rounding` has a default of `none`. If we do need to change it, we can add a codemod

PR #248, opened 2023-09-14
Box:  add Overflow, OverflowX and OverflowY props

somethiiing:
There were a few times where I needed columnReverse for mWeb designs. so this is an update for that.

In addition, overflowX and overflowY styling to avoid using `__dangerouslySetInlineStyle` for things like rounding.

### All directions: (including the two new reverse ones)
image

### Overflow:
image


packages/syntax-core/src/Box/Box.tsx
```diff
@@ -60,6 +60,13 @@ type Margin =
   | 12
   | "auto";
 type Padding = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+type Overflow =
```
somethiiing, line 63: @christianvuerings it's possible that I went a little overboard on these. probably only _really_ need `visible` `hidden` `scroll` and `inherit`, but `auto` + `initial` seemed marginally useful as well?

do you have opinions on this?

packages/syntax-core/src/Box/Box.tsx
```diff
@@ -418,6 +453,9 @@ const Box = forwardRef<HTMLDivElement, BoxProps>(function Box(
     lgMarginEnd,
     lgMarginStart,
     lgMarginTop,
+    // Overflow
+    overflowX,
+    overflowY,
```
benjroy, line 458: @somethiiing are you planning to wire in the `smOverflowX`, `lgOverflowX` ... responsive versions of these props in this PR?

somethiiing: i could, but i didnt think `Overflow` was one of those props that needed a sm and lg variation. I could be wrong about this, what do others think?

packages/syntax-core/src/Box/Box.stories.tsx
```diff
@@ -207,6 +220,46 @@ export const BackgroundColor: StoryObj<typeof Box> = {
   ),
 };
 
+export const Direction: StoryObj<typeof Box> = {
```
somethiiing, line 223: this story probably doesnt need to be here, but it's still fairly nice i guess

packages/syntax-core/src/Box/Box.tsx
christianvuerings: @somethiiing can we restrict it to:

```
hidden
scroll
auto
```

packages/syntax-core/src/Box/Box.tsx
```diff
@@ -265,6 +272,48 @@ type BoxProps = {
    * @defaultValue "start"
    */
   justifyContent?: JustifyContent;
+  /**
+   * How box behaves when content does not fit in the box on both axes.
+   *
+   * Responsive props:
```
christianvuerings, line 278: @somethiiing these are not responsive for now

christianvuerings (approved):

PR #250, opened 2023-09-18
Feat: IconButton - color="black"

benjroy:
## What

This adds `"black"` as a valid `Color` and wires it in so that `IconButton` can display a black button

## Why

A design I am working from is wanting to use a round button that is black:

image

### Story

There is a new story in this PR that renders `IconButton` using all `color` prop values that it accepts:  As of now that looks like this:

image

### Notes

`IconButton` isn't the exact component I need to implement that design (locall in my feature branch I have a very ugly hack that is passing in a fake `icon` componenttype that can render text `1`,`2`,`3`,`4`, etc... inside instead of an Icon

In a separate PR I am going to be proposing a new `RoundButton` component (I will take absolutely any and all suggestions for naming it!) that will allow the round button design used in IconButton to be used to render a button of fixed size with text children.  That is not this PR.

### DIFF
--- a/.changeset/gold-knives-live.md
+++ b/.changeset/gold-knives-live.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+IconButton: add "black" as valid color prop value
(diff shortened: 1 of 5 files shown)


benjroy:
Closing for now

PR #255, opened 2023-09-21
Layer: fix document is not defined on SSR

christianvuerings:
# Changes

* Fixes `document is not defined` when server rendering `Layer`
* Add tests

# Why

@emmanuelgamboa09 raised the issue:
image
https://cambly.slack.com/archives/C04QEEZMSLF/p1695327506767059

# Test

First time we add a server test in Syntax, we should probably do this for other components as well

Failure in the test if we don't fix the issue:
Screenshot 2023-09-21 at 4 50 57 PM

### DIFF
--- a/.changeset/funny-dolls-know.md
+++ b/.changeset/funny-dolls-know.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+Layer: fix document is not defined issue
--- a/packages/syntax-core/src/Modal/Layer.server.test.tsx
+++ b/packages/syntax-core/src/Modal/Layer.server.test.tsx
@@ -1,0 +1,19 @@
+/**
+ * @vitest-environment node
+ */
+import Layer from "./Layer";
+import { renderToString } from "react-dom/server";
+
+describe("layer - server", () => {
+  it("renders on the server without any errors", () => {
+    const renderOnServer = () =>
+      renderToString(
+        <Layer>
+          <div data-testid="child">Content</div>
+        </Layer>,
+      );
+
+    expect(renderOnServer).not.toThrow();
+    expect(renderOnServer()).toStrictEqual("");
+  });
+});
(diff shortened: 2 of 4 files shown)


emmanuelgamboa09 (approved):

somethiiing (approved):

benjroy (approved):
Aside:

Is there no way we can render into a portal during SSR? Having  the modal html being in the server response (but visually hidden) is a requirement for accessibility/screen readers iirc.


This question should not block this bugfix merge, but longer term our syntax components should be able to fully render all their content during SSR

(also, iirc, if/when we adopt ariakit, we can implement our modal on top of ariakit Dialog and just get all of that for free)

christianvuerings:
> Aside:
> 
> Is there no way we can render into a portal during SSR? Having the modal html being in the server response (but visually hidden) is a requirement for accessibility/screen readers iirc.
> 
> This question should not block this bugfix merge, but longer term our syntax components should be able to fully render all their content during SSR
> 
> (also, iirc, if/when we adopt ariakit, we can implement our modal on top of ariakit Dialog and just get all of that for free)

@benjroy usually modals only show up after a client interaction (e.g. `onClick`) so I'm not sure whether this would be needed. Also, React portal relies on a DOM node, so I'm not sure how you would reference that on the server: https://react.dev/reference/react-dom/createPortal

And I just tried SSR rendering `Ariakit.Dialog` and it never shows the dialog contents in the DOM, not even when `open ` is set to true: https://gist.github.com/christianvuerings/b3fcf3ae09c79c250cb2d79566c669b0

results in:

```
- Expected
+ Received

+ <span id=":R0:" style="position:fixed" hidden=""></span>
```

diegohaz:
> And I just tried SSR rendering `Ariakit.Dialog` and it never shows the dialog contents in the DOM, not even when `open ` is set to true: https://gist.github.com/christianvuerings/b3fcf3ae09c79c250cb2d79566c669b0

Hi! To render the Ariakit Dialog open on the server, you must disable the portal, as `ReactDOM.createPortal` doesn't work on the server:
```jsx
<Dialog portal={false}>
```

You can see it working on the server on the [Dialog with Next.js App Router](https://ariakit.org/examples/dialog-next-router) example. Check out the `@login/login/page.tsx` file.

PR #258, opened 2023-10-05
Card: Deprecate 'size' prop, always stretch to width of container

zprong:
# Changes

* Deprecate `size` prop
* Make width stretch to size of container

# Why

* We currently lack a dynamic way of rendering the size past 744px. This blocks us from using this component to render the following design

Existing implementation:
image

Ideal implementation:

image


packages/syntax-core/src/Card/Card.tsx
```diff
@@ -28,7 +28,7 @@ type CardType = {
  */
 export default function Card({
   children,
-  size = "sm",
+  size,
```
somethiiing, line 31: do you want to get rid of the size prop entirely and have devs just manually input maxWidth?

zprong: Hmmm I'm not that opposed to it, but this `size` interface still feels simpler to me

packages/syntax-core/src/Card/Card.tsx
```diff
@@ -18,7 +18,7 @@ type CardType = {
    * `sm`: 352px
    * `lg`: 744px
    *
-   * @defaultValue sm
+   * @defaultValue `undefined`
```
christianvuerings, line 21: @zprong pending slack discussion but could we add `@deprecated` to `size`?

Just make sure we do update the places in our code where we do use `<Card />` without a `size` prop at the moment

packages/syntax-core/src/Card/Card.tsx
```diff
@@ -41,7 +41,7 @@ export default function Card({
       rounding="xl"
       padding={7}
       smPadding={9}
-      maxWidth={sizeWidth[size]}
+      maxWidth={size ? sizeWidth[size] : undefined}
```
christianvuerings, line 44: @zprong a bit shorter perhaps: `maxWidth={size && sizeWidth[size]}`

packages/syntax-core/src/Card/Card.tsx
zprong: ^ For posterity, we'll just make the changes to replace all instances of `size` with `maxWidth` instead of marking as `@deprecated`

zprong:
@somethiiing @christianvuerings please review at your earliest convenient :) 

Draft PR for updates on Cambly-Frontend can be found [here](https://github.com/Cambly/Cambly-Frontend/pull/6922)

somethiiing (approved):
pr looks great. if @christianvuerings is cool with this approach rather than always just having it stretch to `100%` and having the outside container control it, then this should be good to merge.

PR #262, opened 2023-10-17
Interactive components: enable after hydration completes

christianvuerings:
# Context

When any of our interactive elements render it seems like they are immediately intractable, however since hydration hasn't happened yet, the initial state might get lost.

# Current solution

We toggle the hydration state globally and use that in our tests
https://github.com/Cambly/Cambly-Frontend/pull/6943

## Downsides of current solutions

1. Doesn't solve the underlying issue - if a user immediately interacts without hydration being completed, we'll lose the initial state.
2. Requires extra hydration checks in all of our integration tests:
    ```tsx
    await page.waitForSelector("[data-hydrated]");
    await page.getByTestId("subscribe-button").click();
    ```

# Design impact

1. first render a disabled interactive element
2. load all the JavaScript
3. enable the interactive element once the JavaScript finished loading & running

# Resources

## Playwright docs

https://playwright.dev/docs/navigations#hydration

> At some point in time, you'll stumble upon a use case where Playwright performs an action, but nothing seemingly happens. Or you enter some text into the input field and it will disappear. The most probable reason behind that is a poor page [hydration](https://en.wikipedia.org/wiki/Hydration_(web_development)).

> When page is hydrated, first, a static version of the page is sent to the browser. Then the dynamic part is sent and the page becomes "live". As a very fast user, Playwright will start interacting with the page the moment it sees it. And if the button on a page is enabled, but the listeners have not yet been added, Playwright will do its job, but the click won't have any effect.

> A simple way to verify if your page suffers from a poor hydration is to open Chrome DevTools, pick "Slow 3G" network emulation in the Network panel and reload the page. Once you see the element of interest, interact with it. You'll see that the button clicks will be ignored and the entered text will be reset by the subsequent page load code. The right fix for this issue is to make sure that all the interactive controls are disabled until after the hydration, when the page is fully functional.

## React docs
https://react.dev/reference/react/useLayoutEffect#im-getting-an-error-uselayouteffect-does-nothing-on-the-server

> Alternatively, you can render a component with useLayoutEffect only after hydration. Keep a boolean isMounted state that’s initialized to false, and set it to true inside a useEffect call. Your rendering logic can then be like return isMounted ? <RealContent /> : <FallbackContent />. On the server and during the hydration, the user will see FallbackContent which should not call useLayoutEffect. Then React will replace it with RealContent which runs on the client only and can include useLayoutEffect calls.

### DIFF
--- a/packages/syntax-core/src/IconButton/IconButton.tsx
+++ b/packages/syntax-core/src/IconButton/IconButton.tsx
@@ -1,9 +1,12 @@
+"use client";
+
 import classNames from "classnames";
 import backgroundColor from "../colors//backgroundColor";
 import foregroundColor from "../colors/foregroundColor";
 import React, { forwardRef } from "react";
 import { type Color, type Size } from "../constants";
 import styles from "./IconButton.module.css";
+import useIsHydrated from "../useIsHydrated";
 
 const iconSize = {
   ["sm"]: styles.smIcon,
@@ -73,13 +76,15 @@ const IconButton = forwardRef<HTMLButtonElement, IconButtonProps>(
     }: IconButtonProps,
     ref,
   ) => {
+    const isHydrated = useIsHydrated();
+
     return (
       <button
         aria-label={accessibilityLabel}
         data-testid={dataTestId}
         type="button"
         title={tooltip}
-        disabled={disabled}
+        disabled={!isHydrated || disabled}
         onClick={onClick}
         className={classNames(
           styles.iconButton,
--- a/.changeset/cold-bobcats-chew.md
+++ b/.changeset/cold-bobcats-chew.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+Interactive elements: only enable after hydration
--- a/packages/syntax-core/src/Button/Button.server.test.tsx
+++ b/packages/syntax-core/src/Button/Button.server.test.tsx
@@ -1,0 +1,21 @@
+/**
+ * @vitest-environment node
+ */
+import Button from "./Button";
+import { renderToString } from "react-dom/server";
+
+describe("button - server", () => {
+  it("renders on the server without any errors & is disabled before hydration", () => {
+    const renderOnServer = () =>
+      renderToString(
+        <Button
+          accessibilityLabel="Continue to the next step"
+          text="Continue"
+        ></Button>,
+      );
+
+    expect(renderOnServer).not.toThrow();
+    expect(renderOnServer()).toContain("disabled");
+    expect(renderOnServer()).toContain("Continue to the next step");
+  });
+});
--- a/packages/syntax-core/src/Button/Button.tsx
+++ b/packages/syntax-core/src/Button/Button.tsx
@@ -12,6 +12,7 @@ import iconSize from "./constants/iconSize";
 import textVariant from "./constants/textVariant";
 import loadingIconSize from "./constants/loadingIconSize";
 import styles from "./Button.module.css";
+import useIsHydrated from "../useIsHydrated";
 
 type ButtonProps = {
   /**
@@ -117,6 +118,8 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(
     }: ButtonProps,
     ref,
   ) => {
+    const isHydrated = useIsHydrated();
+
     return (
       <button
         data-testid={dataTestId}
@@ -124,7 +127,7 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(
         aria-label={accessibilityLabel}
         type={type}
         title={tooltip}
-        disabled={disabled || loading}
+        disabled={!isHydrated || disabled || loading}
         onClick={onClick}
         className={classNames(
           styles.button,
--- a/packages/syntax-core/src/Checkbox/Checkbox.server.test.tsx
+++ b/packages/syntax-core/src/Checkbox/Checkbox.server.test.tsx
@@ -1,0 +1,24 @@
+/**
+ * @vitest-environment node
+ */
+import Checkbox from "./Checkbox";
+import { renderToString } from "react-dom/server";
+
+describe("checkbox - server", () => {
+  it("renders on the server without any errors & is disabled before hydration", () => {
+    const renderOnServer = () =>
+      renderToString(
+        <Checkbox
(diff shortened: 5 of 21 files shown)


packages/syntax-core/src/IconButton/IconButton.tsx
```diff
@@ -73,13 +76,15 @@ const IconButton = forwardRef<HTMLButtonElement, IconButtonProps>(
...
+    const isMounted = useIsMounted();
+
     return (
       <button
         aria-label={accessibilityLabel}
         data-testid={dataTestId}
         type="button"
         title={tooltip}
-        disabled={disabled}
+        disabled={!isMounted.current || disabled}
```
benjroy, line 87: What do you think the implications of this change will be re: SEO robots?  It seems like the impact from this change specifically would be small, regardless.  That said, if this behavior needs to expand to other Buttons, would returning all of our buttons with `disabled` in the html response from the server affect the way an SEO robot sees/scores our pages?

It _seems_ like it won't be a big issue because these are buttons, and not links with `href` tags, so their behavior is dependent on js runtime behavior.  I just want to raise the concept for us to ponder a moment before going forward with this approach.

packages/syntax-core/src/useIsMounted.tsx
```diff
@@ -0,0 +1,13 @@
...
+  const isMounted = useRef(false);
+
+  useEffect(() => {
+    isMounted.current = true;
+    return () => {
+      isMounted.current = false;
+    };
+  }, []);
+  return isMounted;
+}
```
benjroy, line 13: Naming nit, not blocking:

do you return `isMounted` ref directly because returning `isMounted.current` could have a stale value by the time it needs to be used?

If so, I think this hook should be named `useIsMountedRef`

```suggestion
export default function useIsMountedRef(): React.MutableRefObject<boolean> {
  const isMounted = useRef(false);

  useEffect(() => {
    isMounted.current = true;
    return () => {
      isMounted.current = false;
    };
  }, []);
  return isMounted;
}
```

Otherwise, for naming clarity, it feels like this hook should return `isMounted.current` directly.  What do you think?

```suggestion
export default function useIsMounted(): React.MutableRefObject<boolean> {
  const isMounted = useRef(false);

  useEffect(() => {
    isMounted.current = true;
    return () => {
      isMounted.current = false;
    };
  }, []);
  return isMounted.current;
}
```

benjroy (commented):
Looks okay.  Holding off on the ✅ to give a moment to reflect on the SEO question

packages/syntax-core/src/IconButton/IconButton.tsx
christianvuerings: @christianvuerings follow up and use Google pagerank tools - SEO

christianvuerings: @benjroy seems like crawlers like Google bot do not click buttons on the page, only links:
* https://support.google.com/webmasters/thread/40990043/does-google-bot-click-buttons?hl=en#:~:text=They%20don't%20click%20but,they%20won't%20crawl%20it.
* https://obrienmedia.co.uk/seo/googlebot-does-not-click-on-buttons-how-does-this-affect-seo

I am keeping any `ButtonLink` changes out of this PR but it seem like a possible solution would be to use `pointer-events: none` on the link. That way GoogleBot will still craw the link but the user can't click it until hydration completed
* https://github.com/microsoft/playwright/issues/12298

packages/syntax-core/src/useIsMounted.tsx
christianvuerings: @benjroy @somethiiing changed this to an `isHydrated` hook which just returns a boolean. Feel free to take a look at other implementations:
* https://github.com/search?q=useIsHydrated&type=code 
* https://github.com/jmlweb/isMounted/blob/master/index.es.js
* https://gist.github.com/jaydenseric/a67cfb1b809b1b789daa17dfe6f83daa 
* https://github.com/webNeat/react-tidy/tree/master/src/useIsMounted 
* https://github.com/metric-space-ai/octopus_client/blob/c34583448cfc901e235bbf5837178abf25dd56b7/src/hooks/useIsHydrated.tsx#L10 

As for the previous questions, why not just return `isMounted.current` because in that case, it would be required to pass it as a dependency in any `useEffect`:
* https://www.codemzy.com/blog/ismounted-hook-with-useeffect-reactjs#adding-usecallback-to-create-a-custom-hook 
* https://github.com/jmlweb/isMounted/issues/1

Going to mark this as resolved since we no longer use a ref and just return a boolean

christianvuerings:
@benjroy @somethiiing one more ping, could either of you take another look?

packages/syntax-core/src/IconButton/IconButton.tsx
benjroy: > I am keeping any ButtonLink changes out of this PR but it seem like a possible solution would be to use pointer-events: none on the link.

Should we make that behavior consistent across the different components that need to look at this?  I.e. if we put something in that does `pointer-events: none`, could that same behavior be applied and work for all the components in this PR that are currently getting changes to the their `disabled` prop?

If so, then maybe we should go with that direction.  It seems like the solution could be applied more generically if that were the case?

somethiiing (approved):

packages/syntax-core/src/IconButton/IconButton.tsx
```diff
@@ -1,9 +1,12 @@
+"use client";
+
```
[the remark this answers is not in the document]
christianvuerings: @benjroy we should start adding `"use client"` to all of our components which are interactive.
See https://github.com/mui/material-ui/pull/37656 for more info

`"use client"` means that the component can still be rendered on the server but they will also be included in our JS bundles so that they can re-render on the client.

[our JS bundles](https://www.joshwcomeau.com/react/server-components/#specifying-client-components-5)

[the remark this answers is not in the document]
benjroy: Is this the only file in the repo that needs this, or is this absolutely necessary to have on this file in this PR in order for these changes to work?

If it is not, maybe a separate PR would be more appropriate so that we don't miss anything in the context of this statement:

> we should start adding "use client" to all of our components which are interactive.

benjroy (approved):

PR #264, opened 2023-10-19
[SYN-13] Card styling updates + Box rounding change

somethiiing:
Also manually showed this to @ajmooo and he signed off on all of these

## Card

- update padding from 36px -> 28px on md + lg breakpoints
- update padding from 28px -> 20px on sm breakpoint

## Box

- update `xl` rounding from 32px to 24px

image
image

### DIFF
--- a/.changeset/famous-lizards-melt.md
+++ b/.changeset/famous-lizards-melt.md
@@ -1,0 +1,12 @@
+---
+"@cambly/syntax-core": minor
+---
+
+## Card
+
+- update padding from 36px -> 28px on md + lg breakpoints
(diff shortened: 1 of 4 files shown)


benjroy (approved):
this looks good.  Before merging, can you tell us if you see many spots in the frontend app where border-radius is manually applied that will need to be updated to match this change?

somethiiing:
> this looks good. Before merging, can you tell us if you see many spots in the frontend app where border-radius is manually applied that will need to be updated to match this change?

From my quick grep I think just 1, I can update it as part of https://github.com/Cambly/Cambly-Frontend/pull/7070

PR #265, opened 2023-10-19
MiniActionCard: Remove MiniActionCard from Syntax

somethiiing:
# MiniActionCard: Remove from Syntax

MiniActionCard was the first component added to Syntax and was meant to be a "test" component. As it isn't used much and we no longer need the basic demonstration, it can be removed.

The [PR](https://github.com/Cambly/Cambly-Frontend/pull/7070) to add/replace Syntax MiniActionCard is up. Will need to merge that before bumping _this_ version of Syntax


christianvuerings (approved):
@somethiiing thanks for doing this! Please be sure to first land the Cambly-Frontend PR (which removes the usage) before landing this one

.changeset/modern-islands-cheer.md
```diff
@@ -0,0 +1,7 @@
+---
+"@cambly/syntax-core": major
```
benjroy, line 2: Do we have any other breaking changes lined up that we'd like to include in this bump?

benjroy (approved):

PR #270, opened 2023-10-25
FIX: Checkbox: Fix input overlay rendering bug that forced way too much extra height in overflow containers

benjroy:
Been looking into some tricky rendering issues with @kpatelPro over in our frontend repo.  after a long dive, we were able to trace the issue down to here.

Here's a video that of shows the issue with the storybook example I added with this PR, before the change in `Checkbox.module.css`

https://github.com/Cambly/syntax/assets/1890801/2896e7ba-f5a8-416f-85b3-7ad5a5f3f56c

I'm going to let @kpatelPro add more description in with this if he would like.

### DIFF
--- a/packages/syntax-core/src/Checkbox/Checkbox.stories.tsx
+++ b/packages/syntax-core/src/Checkbox/Checkbox.stories.tsx
@@ -53,7 +53,7 @@ export const Interactive: StoryObj<typeof Checkbox> = {
 
 export const FixDoesNotBlowoutHeight: StoryObj<typeof Checkbox> = {
   render: () => (
-    <Box height="100px" overflowY="auto">
+    <Box height="100px" overflowY="auto" padding={3}>
       <CheckboxInteractive />
       <CheckboxInteractive />
       <CheckboxInteractive />
--- a/packages/syntax-core/src/Checkbox/Checkbox.stories.tsx
+++ b/packages/syntax-core/src/Checkbox/Checkbox.stories.tsx
@@ -1,6 +1,7 @@
 import { type StoryObj, type Meta } from "@storybook/react";
 import Checkbox from "./Checkbox";
 import React, { useState } from "react";
+import Box from "../Box/Box";
 
 export default {
   title: "Components/Checkbox",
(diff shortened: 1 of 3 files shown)


kpatelPro:
Long-story, short: A container with `Checkbox` descendants appeared to be exceeding  its declared height and forcing an ancestor container to manifest a scrollbar.  This was because `Checkbox` was including element `inputOverlay` with style `position: absolute` without including a corresponding container element with style `position: relative`, so the `inputOverlay` elements were stacking outside the bounds of the `Checkbox` itself.

kpatelPro (approved):
LG.  Thank you!

somethiiing (approved):

packages/syntax-core/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -49,3 +50,31 @@ const CheckboxInteractive = () => {
 export const Interactive: StoryObj<typeof Checkbox> = {
   render: () => <CheckboxInteractive />,
 };
+
+export const FixDoesNotBlowoutHeight: StoryObj<typeof Checkbox> = {
+  render: () => (
+    <Box height="100px" overflowY="auto">
```
christianvuerings, line 56: image
could we add some paddingX / Y to the container? Right now it breaks the focus styles

christianvuerings (approved):
@benjroy thanks for fixing! We should have checked Checkbox when we fixed RadioButton: https://github.com/Cambly/syntax/pull/149

packages/syntax-core/src/Checkbox/Checkbox.stories.tsx
benjroy: To the container in the stories?   Is this unique to this story, or do all the focus styles brush up against the container like in this story?

I'll check.  if it is unique to this story, I'll add in some padding in the example.  If it is not unique to this one, then it'd be better I think to do that in a deparate PR for all of the stories. 👍

christianvuerings: @benjroy Yes, it's unique to this story because of the `overflow` 
image

PR #276, opened 2023-11-15
TapArea: support hover on children with color

christianvuerings:
# Context

https://cambly.slack.com/archives/C033ZPY5M46/p1700070891891909?thread_ts=1699987244.953739&cid=C033ZPY5M46

Previously, when we hovered a TapArea which had a child with color, it would not show the TapArea hover style `rgba(0,0,0,0.1)`. The issue is that both the `Box` and `TapArea` components are trying to set the background.

Fixed it by adding a conditional child element which has the overlay style.

## Before

https://github.com/Cambly/syntax/assets/127199/a3eae106-ebbe-441b-add4-c1dd93acda43

## After

https://github.com/Cambly/syntax/assets/127199/01f7fad9-4520-4999-b3e5-c536caf5fcab

### DIFF
--- a/.changeset/pretty-moose-punch.md
+++ b/.changeset/pretty-moose-punch.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+TapArea: support hover on children with color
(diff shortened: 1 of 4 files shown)


ajmooo (approved):
lgtm

zprong (approved):
Looks great, thanks so much for updating this!!

PR #278, opened 2023-11-20
Add line clamp to heading component

nathan-ong:
Adding a lineClamp prop


christianvuerings (approved):
looks good - don't forget to add a changeset (minor)

PR #281, opened 2023-11-22
TapArea: fix issue with unnecessary focus

zprong:
There's an issue with TapArea being focused unnecessarily when clicked discussed in [this slack thread](https://cambly.slack.com/archives/C033ZPY5M46/p1700511311033089)

https://github.com/Cambly/syntax/assets/36240753/16dc28fc-2f97-49e9-91eb-9755a72918b4

### DIFF
--- a/.changeset/perfect-queens-turn.md
+++ b/.changeset/perfect-queens-turn.md
@@ -2,4 +2,4 @@
 "@cambly/syntax-core": minor
 ---
 
-Don't consider TapArea as "focussed" if clicked
+Blur TapArea event target when clicked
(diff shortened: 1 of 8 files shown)


christianvuerings:
@zprong there seems to be an issue with the actual focus state - somehow all TapArea's show the focus style now when 1 is focussed. Might be a bug with `useFocusVisible ` itself. For now it might be ok to call `.blur()` on `event.currentTarget` when clicking the `TapArea`

https://github.com/Cambly/syntax/assets/127199/bda68737-79a4-4e3e-acec-fe7c4686b6ce

zprong:
Oh no! Thank you for catching that @christianvuerings ! I'll do that instead

zprong:
@christianvuerings I've confirmed that this is no longer an issue with using `blur()` instead!

https://github.com/Cambly/syntax/assets/36240753/24b5de5e-b0f7-4a66-be23-6eb0db5e43c6

PR #249, opened 2023-09-18
tree-shaking: build individual module files instead of single index.js entry bundle

benjroy:
## What

builds/transpiles individual modules into `dist` instead of a single `index.js` bundle file.

## Why

This should allow for better tree-shaking when bundling apps

#### Notes:

The [log for the build job](https://github.com/Cambly/syntax/actions/runs/6218519342/job/16875104418?pr=249) is the most interesting artifact to inspect in this PR:

<details>

```
  pnpm run build
  shell: /usr/bin/bash -e {0}
  env:
    PNPM_HOME: /home/runner/setup-pnpm/node_modules/.bin
    NODE_OPTIONS: --max-old-space-size=8192

> @ build /home/runner/work/syntax/syntax
> turbo run build

• Packages in scope: @cambly/eslint-config-syntax, @cambly/syntax-codemods, @cambly/syntax-core, @cambly/syntax-design-tokens, @cambly/syntax-floating-components, @cambly/syntax-tsconfig, @cambly/syntax-utils, @syntax/storybook
• Running build in 8 packages
• Remote caching disabled
@cambly/syntax-design-tokens:build: cache miss, executing 9b5294ab3acc124a
@cambly/syntax-codemods:build: cache miss, executing c5641a1d87ef0098
@cambly/syntax-utils:build: cache miss, executing 5e30b24b2f3cac33
@cambly/syntax-design-tokens:build: 
@cambly/syntax-design-tokens:build: > @cambly/syntax-design-tokens@0.9.0 build /home/runner/work/syntax/syntax/packages/syntax-design-tokens
@cambly/syntax-design-tokens:build: > rm -rf dist;style-dictionary build
@cambly/syntax-design-tokens:build: 
@cambly/syntax-codemods:build: 
@cambly/syntax-codemods:build: > @cambly/syntax-codemods@0.5.1 build /home/runner/work/syntax/syntax/packages/syntax-codemods
@cambly/syntax-codemods:build: > tsup
@cambly/syntax-codemods:build: 
@cambly/syntax-utils:build: 
@cambly/syntax-utils:build: > @cambly/syntax-utils@1.3.0 build /home/runner/work/syntax/syntax/packages/syntax-utils
@cambly/syntax-utils:build: > tsup src/index.tsx --format esm,cjs --dts --external react
@cambly/syntax-utils:build: 
@cambly/syntax-design-tokens:build: 
@cambly/syntax-design-tokens:build: scss
[...]

PR #287, opened 2023-12-13
20231212 scripts for auto write package json exports

benjroy:
Drafting this for now, will provide full description when ready

Working on followup from https://github.com/Cambly/syntax/pull/249#issuecomment-1853043353

> > Ideally we would be stricter as to what folks would be able to import - e.g. right now we're also exporting packages/syntax-core/dist/colors/allColors.js and folks would be able to import that, which is not ideal. We can limit these by using exports in package.json: https://github.com/typescript-eslint/typescript-eslint/blob/2a2a76ad6540177af19cbb4194895496978d288a/packages/utils/package.json#L13-L39
> 
> I have a branch that autowrites the export fields in package.json. It involves a lot of scripting that I lifted/modified from the ariakit repo's build scripts. We can use that PR to restrict down which paths will be importable. I'm in this branch locally at the moment catching it up.


tsup.config.ts
```diff
@@ -23,6 +24,7 @@ export default defineConfig({
   sourcemap: true,
   dts: true,
   splitting: true,
+  bundle: false,
```
benjroy, line 27: 👀 keep an eye on this: re building css modules

PR #290, opened 2023-12-25
Add dependency review GitHub action

christianvuerings:
Enables https://github.com/actions/dependency-review-action to ensure we don't add new security vulnerabilities or packages with invalid licenses

Note that we will only be able to enable this on our private repos once we have [GitHub advanced security](https://docs.github.com/en/get-started/learning-about-github/about-github-advanced-security) enabled
> [$49 per month per active committer.](https://resources.github.com/contact/security)

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,17 @@
+name: "Dependency Review"
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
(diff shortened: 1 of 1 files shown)


somethiiing (approved):
do you have a eta on when we have approval for getting advanced security?

christianvuerings:
> do you have a eta on when we have approval for getting advanced security?

No ETA, just would mean a pretty significant cost increase

PR #292, opened 2024-01-09
Set `"type": "module"` in bundle-able packages' package.json files

benjroy:
## Background:

I've been exploring the differences in building components on top of `ariakit` vs `react-aria`.  I have branches with implementations of a `Tooltip` component shipped from `syntax-core`.  I've been locally building and packing (`pnpm run build && pnpm pack`) syntax-core and installing it in our frontend repo to build and inspect bundle-size impacts.  Here's the rudimentary stats I gathered:

```
Before: no tooltip on kids courses page:                                                                11.8 kB         282 kB
Before: syntax-floating-components Tooltip (v1) added to kids/courses page:   11.9 kB         298 kB
After: react-aria syntax-core Tooltip (v2) added to kids/courses page:               225 kB         495 kB
After: ariakit syntax-core Tooltip (v2) added to kids/courses page:                    84.3 kB         355 kB

 
Base:  nothing rendered on kids/courses page.  No tooltip no nothing                 1.88 kB         266 kB
Before: syntax-floating-components Tooltip (v1) only child kids/courses page:   1.97 kB         282 kB
After: react-aria syntax-core Tooltip (v2) only child kids/courses page:                215 kB         480 kB
After: ariakit syntax-core Tooltip (v2) only child kids/courses page:                     74.4 kB         339 kB


(Ah this is important)
With `"type": "module"` set in syntax-core package.json:

Base: no tooltip no nothing on kids/courses page:                                             1.88 kB         266 kB
After: react-aria syntax-core Tooltip (v2) only child kids/courses page:           28.3 kB         293 kB
After: ariakit syntax-core Tooltip (v2) only child kids/courses page:                 24.9 kB         289 kB
```

I was getting errors importing the `ariakit` versions that hinted that it was picking the wrong built files, so I tried out the `"type": "module"` change in the package.json to see if that made the nextjs build errors go away.

Not only did it fix that issue, it had a very significant impact on the bundle-size additions when syntax ships either `ariakit` or `react-aria` as a dependency.  Thus this PR:

# Changes

This PR adds `type: "module"` to the package.json files for packages that our express and next apps build into their bundles.

Ensures that ESM modules are picked for dependencies that syntax packages are shipping when syntax packages get built into top-level apps.

Patch version bumps.


benjroy:
lint run is failing for reasons like this:

```
@cambly/syntax-core:lint: Error [ERR_REQUIRE_ESM]: require() of ES Module /home/runner/work/syntax/syntax/packages/syntax-core/.eslintrc.js from /home/runner/work/syntax/syntax/node_modules/@eslint/eslintrc/dist/eslintrc.cjs not supported.
@cambly/syntax-core:lint: .eslintrc.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
@cambly/syntax-core:lint: Instead rename .eslintrc.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /home/runner/work/syntax/syntax/packages/syntax-core/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

.... I was running into this issue when working last week in a branch to tighten up the importable paths/auto-write the exports fields in package.json files.  I punted on that task for now because of the eslint issues getting in the way, but this PR might be forcing-function to finally deal with that issue :+1:

christianvuerings:
Initial PR looks good - let's indeed figure out how to fix the lint issues before merging

christianvuerings:
@benjroy this should be fixed now. Just be sure that when we land, we immediately bump the version in Cambly-Frontend.

christianvuerings (approved):

benjroy:
Thanks @christianvuerings!  I'm going to wait until the react-aria popover/tooltip implementations are ready/up for review before merging this.

PR #294, opened 2024-01-10
TextArea: add component & update props

christianvuerings:
# Description

Adds `TextArea` component which previously lived in `Cambly-Frontend`

# Changes

- Props: Add `data-testid`, `errorText`, `helperText`, `size` and `id` props
- Props: Make `maxLength` configurable and set default to 1024; 
- Props: Make `placeholder` optional, just as with `TextField`
- Test: Add client and server tests
- Documentation: Document each prop
- SSR support: Support Server Side Rendering by disabling the element before hydration

Also updated `TextField` so the font size + padding changes depending on the `size` - as defined in Figma.

# Screenshots

Screenshot 2024-01-10 at 5 08 09 AM
Screenshot 2024-01-10 at 5 08 02 AM

### DIFF
--- a/.changeset/wise-apples-invite.md
+++ b/.changeset/wise-apples-invite.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+TextArea: add component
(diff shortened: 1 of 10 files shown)


benjroy (approved):
Looks good to me ✅ 🚢   Thanks for finally hoisting this up here from the FE repo!

somethiiing (approved):

PR #295, opened 2024-01-11
AvatarGroup: update AvatarGroup API using context

somethiiing:
- Updated AvatarGroup using React.Context 
- Update styling + icon position on Avatar

The default Avatar Icon has a thin border around it, so it makes it look like the borders are uneven for the ones that have default vs the filled one, but they are correct

image


packages/syntax-core/src/Avatar/Avatar.module.css
```diff
@@ -1,31 +1,32 @@
...
 
-.avatar {
+.avatarImage {
+  box-sizing: border-box;
   background-position: center;
   background-size: cover;
   background-repeat: no-repeat;
   border-radius: 50%;
-  border: var(--syntax-avatar-border-width) solid #fff;
+  border: 2px solid #fff;
```
christianvuerings, line 11: Note that this changes the border width from 1px to 2px @ajmooo

packages/syntax-core/src/Avatar/Avatar.tsx
```diff
@@ -1,14 +1,66 @@
+"use client";
```
christianvuerings, line 1: @somethiiing Could we remove this for now? This will break once we try to use it in Cambly-Frontend https://github.com/Cambly/syntax/pull/274

packages/syntax-core/src/AvatarGroup/AvatarGroup.tsx
```diff
@@ -1,40 +1,85 @@
...
+          },
+        }}
+      >
+        {children}
+      </Box>
+    </AvatarGroupContext.Provider>
   );
 }
-
-AvatarGroup.displayName = "AvatarGroup";
```
christianvuerings, line 39: @somethiiing could we add this back in?

christianvuerings (approved):
@somethiiing thanks for making these fixes! I would verify that we do not use this in a server component before deploying.

PR #313, opened 2024-02-20
Colored cards

nathan-ong:
Add the ability to change the background color on our cards


somethiiing (approved):

packages/syntax-core/src/Card/Card.tsx
```diff
@@ -7,6 +8,10 @@ type CardType = {
    * Test id for the button
    */
   "data-testid"?: string;
+  /**
+   * The background color of the box.
```
christianvuerings, line 12: @nathan-ong could you add a `@defaultValue`?

christianvuerings (approved):

PR #314, opened 2024-02-22
TapArea: add onMouseEnter and onMouseLeave events

christianvuerings:
In order to build custom accessible components we need to be able to access `onMouseEnter` and `onMouseLeave` on `TapArea`


.vscode/extensions.json
```diff
@@ -3,6 +3,6 @@
     "eamodio.gitlens",
     "esbenp.prettier-vscode",
     "dbaeumer.vscode-eslint",
-    "ZixuanChen.vitest-explorer"
```
christianvuerings, line 6: Extension was deprecated

PR #289, opened 2023-12-20
Facepile

nathan-ong:
Figma design:
https://www.figma.com/file/p7LKna9JMU0JEkcKamzs53/%F0%9F%93%90-Syntax?type=design&node-id=8051-10795&mode=design

Screenshot 2023-12-20 at 12 11 53 PM
Screenshot 2023-12-20 at 12 12 04 PM

PR #317, opened 2024-02-24
ThemeProvider: add component & add cambio versions of Button / IconButton & LinkButton

christianvuerings:
# Description

In order to get ready for Cambio, the Cambly redesign, we're adding a "Cambio" theme and convert buttons over to the new design

# Changes

* ThemeProvider: add component
* Storybook: add theme selector
* Button / IconButton / LinkButton: add Cambio versions

# Video

https://github.com/Cambly/syntax/assets/127199/1b5736cc-f6e5-4a0c-89cb-ef296cd425e4


packages/syntax-core/src/Button/constants/color.ts
```diff
@@ -0,0 +1,66 @@
...
+  | "success"
+  | "inverse" {
+  if (color === "quaternary") {
+    return "inverse";
+  } else if (color === "destructive-tertiary") {
+    return "destructive-secondary";
+  } else if (color === "success-primary" || color === "success-secondary") {
+    return "success";
+  }
+  return color;
```
btorndorff, line 31: Is this here so the classic buttons don't break if we convert them to cambio? Is it going to be removed eventually?

btorndorff (approved):
LGTM just had a question

zprong (approved):
LGTM!

packages/syntax-core/src/Button/Button.tsx
```diff
@@ -30,25 +36,46 @@ type ButtonProps = {
   /**
    * The color of the button
    *
+   * Classic only:
```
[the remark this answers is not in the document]
christianvuerings: @zprong updated - `success` is classic only, `success-primary`, `success-secondary` and `success-tertiary` are cambio only

packages/syntax-design-tokens/config.json
```diff
@@ -40,78 +40,6 @@
           "format": "json/flat"
         }
       ]
-    },
-    "android": {
```
christianvuerings, line 44: Removed since they aren't used

turbo.json
```diff
@@ -7,7 +7,7 @@
     },
     "test": {
       "outputs": ["coverage/**"],
-      "dependsOn": []
+      "dependsOn": ["^build"]
```
christianvuerings, line 10: required now since we rely on a build version of our design tokens when running the tests

packages/syntax-core/src/Button/constants/color.ts
christianvuerings: @btorndorff that is correct - added so both themes work with each other. Goal is to remove the `classic` them eventually

package.json
```diff
@@ -26,7 +26,6 @@
     "glob": "10.3.10",
     "jsdom": "22.0.0",
     "prettier": "2.8.8",
-    "storybook-addon-rtl": "^0.5.0",
```
christianvuerings, line 29: @nathan-ong @somethiiing  fyi follow up from
https://github.com/Cambly/syntax/pull/289/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R29

Storybook specific dependencies should go inside the `storybook` app so moved it to `apps/storybook/package.json`

packages/syntax-core/src/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+const ThemeContext = React.createContext<Theme>({
+  themeName: "classic",
+});
+ThemeContext.displayName = "ThemeContext";
+
+const classicToCambioKeyLookup = {
+  "color-base-black": "color-cambio-black",
+  "color-base-destructive-100": "color-cambio-destructive-100",
+  "color-base-destructive-200": undefined, // Deprecated - to be deleted
```
christianvuerings, line 18: @ajmooo fyi still seeing quite a lot of usage of deprecated colors so just falling back to the legacy "classic" colors and will remove in a follow up

PR #330, opened 2024-03-15
RichSelectList: update color and make label optional

jliotta:
This PR adds the following updates for RichSelectList:
- Make the `label` prop optional (similar to what we do for TextField
- Because the label is now optional, I've added a required `accessibilityLabel` prop
- Make the default color for the select box white

Changes needed for [Shoppable Group Lessons](https://www.figma.com/file/e0BhCuouCduwIhnB3D7NQc/Shoppable-Group-Lessons?node-id=170%3A35748&mode=dev) which is the first use case of RichSelectList

### DIFF
--- a/packages/syntax-core/src/RichSelect/RichSelect.module.css
+++ b/packages/syntax-core/src/RichSelect/RichSelect.module.css
@@ -2,3 +2,7 @@
 .richSelectBox:focus-visible {
   outline: none;
 }
+
+.label {
+  cursor: pointer;
+}
--- a/packages/syntax-core/src/RichSelect/RichSelectList.tsx
+++ b/packages/syntax-core/src/RichSelect/RichSelectList.tsx
@@ -42,10 +42,7 @@ const iconSize = {
   lg: 24,
 } as const;
 
-export type RichSelectListProps = Omit<
-  RichSelectBoxProps,
-  "accessibilityLabel"
-> & {
+export type RichSelectListProps = Omit<RichSelectBoxProps, "id"> & {
   /** Test id for the select element */
   "data-testid"?: string;
   /**
@@ -59,10 +56,12 @@ export type RichSelectListProps = Omit<
   errorText?: string;
   /** Text shown below select box */
   helperText?: string;
+  /**
+   * RichSelectList id, if not provided, a unique id will be generated
+   */
+  id?: string;
   /** Text shown above select box */
   label?: string;
-  //** */
-  accessibilityLabel: string;
   /**
    * Text showing in select box if no option has been chosen.
    * There should always have a placeholder unless there is a default option selected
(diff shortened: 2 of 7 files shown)


packages/syntax-core/src/RichSelect/RichSelect.module.css
```diff
@@ -2,3 +2,7 @@
 .richSelectBox:focus-visible {
```
christianvuerings, line 2: @jliotta could you make sure to add a changeset to this PR? `pnpm run changeset`

packages/syntax-core/src/RichSelect/RichSelectList.tsx
```diff
@@ -59,7 +60,9 @@ export type RichSelectListProps = Omit<
   /** Text shown below select box */
   helperText?: string;
   /** Text shown above select box */
-  label: string;
+  label?: string;
+  //** */
```
christianvuerings, line 64: @jliotta instead of adding new `accessibilityLabel` prop, can we allow for an `id` to be passed in? This follows the same API as `TextField`:

https://github.com/Cambly/syntax/blob/09370d918f3fbb34cab9307ab318e68e03e874f2/packages/syntax-core/src/TextField/TextField.tsx#L106

The way folks can then add a label, is by adding a custom label outside of RichSelectList: https://github.com/Cambly/syntax/blob/09370d918f3fbb34cab9307ab318e68e03e874f2/packages/syntax-core/src/TextField/TextField.tsx#L91-L92

packages/syntax-core/src/RichSelect/RichSelectList.tsx
```diff
@@ -117,6 +121,7 @@ function RichSelectList(props: RichSelectListProps): ReactElement {
     ...richSelectBoxProps
   } = props;
 
+  const id = useId();
```
christianvuerings, line 124: See https://github.com/Cambly/syntax/blob/09370d918f3fbb34cab9307ab318e68e03e874f2/packages/syntax-core/src/TextField/TextField.tsx#L91-L92 - we want similar logic here

packages/syntax-core/src/SelectList/SelectList.module.css
```diff
@@ -32,6 +32,7 @@
   outline: 0;
   padding: 0 36px 0 12px;
   width: 100%;
+  background-color: var(--color-base-white);
```
christianvuerings, line 35: NIT: can we alphabetically sort this?

somethiiing (approved):

jliotta:
@christianvuerings I think I addressed your comments but please take another look! I'm going to merge this in for now since it's blocking some FE work but I can submit a follow-up PR if needed on Monday

PR #335, opened 2024-03-18
RichSelectList: fix classic & cambio styling

christianvuerings:
# Before

Screenshot 2024-03-18 at 9 58 37 AM

# After
Screenshot 2024-03-18 at 9 43 54 AM
Screenshot 2024-03-18 at 9 43 48 AM

### DIFF
--- a/packages/syntax-core/src/RichSelect/RichSelectList.tsx
+++ b/packages/syntax-core/src/RichSelect/RichSelectList.tsx
@@ -33,6 +33,7 @@ import RichSelectSection from "./RichSelectSection";
 import RichSelectChip from "./RichSelectChip";
 import RichSelectRadioButton from "./RichSelectRadioButton";
 import { useField } from "react-aria";
+import { useTheme } from "../ThemeProvider/ThemeProvider";
 
 const NOOP = () => undefined;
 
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/RichSelect/RichSelectList.tsx
```diff
@@ -61,7 +62,7 @@ export type RichSelectListProps = RichSelectBoxProps & {
    */
   id?: string;
   /** Text shown above select box */
-  label?: string;
+  label: string;
```
christianvuerings, line 65: @jliotta `label` should still be required but it can be an empty string. When it is an empty string, the label won't render

packages/syntax-core/src/RichSelect/RichSelectList.tsx
```diff
@@ -176,31 +178,31 @@ function RichSelectList(props: RichSelectListProps): ReactElement {
       <div
         className={classNames(styles.selectContainer, {
           [styles.opacityOverlay]: disabled,
+          [styles.selectContainerCambio]: themeName === "cambio",
         })}
       >
-        <ReactAriaLabel
```
christianvuerings, line 181: `ReactAriaLabel` renders a `label` and we rendered a `label` inside of it. Fix that

jliotta:
Awesome thank you @christianvuerings !

PR #349, opened 2024-03-21
RichSelectList: Update onClick call

jliotta:
Passing in the onClick prop to the TapArea of RichSelectList never actually gets called since the mouse isn't clicking on that element. To trigger the onClick, it has to be passed into the div at the parent level.


somethiiing (approved):

packages/syntax-core/src/RichSelect/RichSelectList.tsx
```diff
@@ -180,6 +180,7 @@ function RichSelectList(props: RichSelectListProps): ReactElement {
           [styles.opacityOverlay]: disabled,
           [styles.selectContainerCambio]: themeName === "cambio",
         })}
+        onClick={onClick}
```
christianvuerings, line 183: @jliotta this adds an inaccessible `onClick` to `div` - it's never reachable by keyboard. Could we come up with a better solution?

PR #357, opened 2024-03-25
Modal/FocusTrap: useLayoutEffect

christianvuerings:
Better fix than https://github.com/Cambly/syntax/pull/355

In Cambio we use a Cambly Syntax IconButton instead of a custom `button` component. An `IconButton` is only enabled after hydration / first render so focus on the first element is not possible which caused the previous `.focus` failure.

How is this better?
Still sets the `focus` on the close icon button

image

### DIFF
--- a/.changeset/tasty-clocks-rule.md
+++ b/.changeset/tasty-clocks-rule.md
@@ -1,0 +1,6 @@
+---
+"@cambly/syntax-core": minor
+"@syntax/storybook": minor
+---
+
+Modal/FocusTrap: useLayoutEffect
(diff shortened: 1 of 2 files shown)


somethiiing (approved):

btorndorff (approved):

christianvuerings:
Closing this PR since it was fixed by using `isHydrated` instead: https://github.com/Cambly/syntax/pull/367/files#diff-8e3a1880ee770b6f67c6a11119842e56cde764b91f8ec2433e59c59b557e62e7R73

(the `useLayoutEffect` approach didn't actually work)

PR #364, opened 2024-03-27
[Modal] add scrollable functionality

somethiiing:
Functionality should behave the same until the height of the modal is too tall.

the modal's max height should be (max view height - 64px (32 from top/bottom margin) ).

@ajmooo  signed off on styling +functionality!

### DIFF
--- a/packages/syntax-core/src/Modal/Modal.stories.tsx
+++ b/packages/syntax-core/src/Modal/Modal.stories.tsx
@@ -4,6 +4,7 @@ import { type StoryObj, type Meta } from "@storybook/react";
 import Button from "../Button/Button";
 import Typography from "../Typography/Typography";
 import Modal from "./Modal";
+import Box from "../Box/Box";
 
 export default {
   title: "Components/Modal",
@@ -99,7 +100,7 @@ export const WithImage: StoryObj<typeof Modal> = {
   args: {
     ...Default.args,
     header: "With Image",
-    image: <img src="https://placehold.co/400x200" alt="placeholder image" />,
+    image: <img src="https://placehold.co/600x200" alt="placeholder image" />,
   },
   render: function WithImageExample({ ...args }): JSX.Element {
     const [isOpen, setIsOpen] = useState(false);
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/Modal/Modal.stories.tsx
```diff
@@ -99,7 +100,7 @@ export const WithImage: StoryObj<typeof Modal> = {
   args: {
     ...Default.args,
     header: "With Image",
-    image: <img src="https://placehold.co/400x200" alt="placeholder image" />,
+    image: <img src="https://placehold.co/600x200" alt="placeholder image" />,
```
somethiiing, line 103: just updating image to fit the correct size per cambio update

packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -236,7 +243,11 @@ export default function Modal({
                 >
                   {header}
                 </Heading>
-                <Box marginBottom={themeName === "classic" ? 4 : 0}>
+                <Box
+                  marginBottom={themeName === "classic" ? 4 : 0}
+                  maxHeight={"calc(100vh - 232px)"} // 232x is combined height of header/footer/padding/margin
```
somethiiing, line 248: @christianvuerings i know you're not a fan of these "magic" numbers. but `maxHeight` is required to get this scrollbar to show up properly, and i think this is "exactly" what is we want here.

I don't know a way of doing a dynamic max height other than like this?

somethiiing: but for documentation's sake:
the math is:

48px header height
16px gap between body/header
48px footer
8px gap between body/footer
32 top margin
32 bottom margin
24 top padding
24 bottom padding

packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -183,10 +183,17 @@ export default function Modal({
...
+              marginTop={8}
+              marginBottom={8}
               minWidth={240}
               maxWidth={sizeWidth[themeName === "classic" ? size : "lg"]}
               width="100%"
-              dangerouslySetInlineStyle={{ __style: { overflow: "hidden" } }}
+              dangerouslySetInlineStyle={{
+                __style: {
+                  overflow: "hidden",
+                  maxHeight: "calc(100vh - 64px)",
```
somethiiing, line 194: 100vh minus 32px top and bottom

packages/syntax-core/src/Modal/Modal.tsx
btorndorff: would % work for it? I used it in `LessonV2DetailsModal` but not sure that would work in this context.

```
      <Box
        maxHeight="80%"
        overflowY="auto"
        dangerouslySetInlineStyle={{
          __style: { scrollbarWidth: "none", paddingBottom: "36px" },
        }}
        gap={3}
        direction="column"
        display="flex"
        marginTop={3}
      >
```

christianvuerings: @somethiiing yes, we should be able to do this without this magic number. Main thing is that it's really hard to keep this number updated if any of the components inside of it change.

Please take a look at https://github.com/Cambly/syntax/pull/367 for an alternative approach

christianvuerings (changes requested):

PR #371, opened 2024-04-02
Button/Badge: Cambio updates

christianvuerings:


ajmooo:
I had intentionally, but inconsistently specced success button vs iconbutton in the component tracker. Can you change the success100 color of tertiary buttons/iconbuttons to success300 instead? I will update the tracker to note this.

ajmooo:
With badge, there are some duplicative colors (gray200 and gray370 look to be the same to my eye? same with 900/870.) I think the only call-out there is it might be confusing to the engineer when to use one vs the other when we know some of these names will be deprecated in the future.

christianvuerings:
> With badge, there are some duplicative colors (gray200 and gray370 look to be the same to my eye? same with 900/870.) I think the only call-out there is it might be confusing to the engineer when to use one vs the other when we know some of these names will be deprecated in the future.

Yes, it's still one of the follow ups to remove these deprecated colors from `Badge`

PR #373, opened 2024-04-04
Modal: Update size of Close (X) Icon on Modal from sm to md

somethiiing:
Update Close icon size from sm to md (20x20 to 48x48)


joemanley201 (approved):

christianvuerings:
@somethiiing might still be ok but just now that the heading can now overlap with the IconButton: 
image

PR #381, opened 2024-04-11
[Box] add optional primary border prop to Box

somethiiing:
Talked with @christianvuerings + @ajmooo about how to handle this. we use this border prop fairly often. having a defined style makes it more consistent usage across the app. Also will eliminate usages of `dangerouslySetInlineStyle` just to set border

image

### DIFF
--- a/packages/syntax-core/src/Box/Box.stories.tsx
+++ b/packages/syntax-core/src/Box/Box.stories.tsx
@@ -68,6 +68,10 @@ export default {
       control: { type: "select" },
       options: allColors,
     },
+    border: {
+      control: { type: "radio" },
+      options: ["primary", "none"],
+    },
     dangerouslySetInlineStyle: {
       control: { type: "object" },
     },
@@ -171,7 +175,7 @@ export default {
       control: "text",
     },
     rounding: {
-      options: ["none", "sm", "md", "lg", "xl", "full"],
+      options: ["none", "sm", "md", "full"],
       control: { type: "select" },
     },
     children: {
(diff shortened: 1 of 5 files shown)


packages/syntax-core/src/Box/Box.tsx
```diff
@@ -98,6 +98,14 @@ type BoxProps = {
...
    */
   backgroundColor?: (typeof allColors)[number];
+  /**
+   * The border color of the box.
+   * This is a passthrough prop directly to the Box component.
+   *
+   * Usage example: `border="1px solid var(--color-base-gray-300)"`
+   * @defaultValue `none`
+   */
+  border?: string;
```
somethiiing, line 108: I'm fully aware this is terrible, I was wondering if @christianvuerings had a better way of typing this?

christianvuerings (changes requested):
Let's not add passthrough props like these - makes it impossible to make them consistent across.

Instead, we should work with design and figure out specific variants

packages/syntax-core/src/Box/Box.stories.tsx
```diff
@@ -406,7 +420,7 @@ const roundingLookup = {
 export const Rounding: StoryObj<typeof Box> = {
   render: () => (
     <Box display="flex" gap={4} flexWrap="wrap">
-      {(["sm", "md", "lg", "xl", "full"] as const).map((rounding) => (
```
somethiiing, line 409: removing old variations of rounding on the story

packages/syntax-core/src/Box/Box.stories.tsx
```diff
@@ -244,9 +248,19 @@ export const BackgroundColor: StoryObj<typeof Box> = {
...
+      <Box border="primary" width={240} padding={2}>
+        <Typography>Border Primary</Typography>
+      </Box>
+    </Box>
+  ),
+};
+
 export const Direction: StoryObj<typeof Box> = {
   render: () => (
-    <>
```
somethiiing, line 249: tiny tweak to make the story look a little better

btorndorff (approved):

PR #385, opened 2024-04-15
[Box] add optional border

somethiiing:
Note: since @christianvuerings is out for a week and he requested changes on my other PR (and so merging is blocked), I'm creating a new one so it isn't blocking my feature work. [original PR](https://github.com/Cambly/syntax/pull/381)

Talked with @christianvuerings + @ajmooo about how to handle this. we use this border prop fairly often. having a defined style makes it more consistent usage across the app. Also will eliminate usages of dangerouslySetInlineStyle just to set border

image

### DIFF
--- a/packages/syntax-core/src/Box/Box.stories.tsx
+++ b/packages/syntax-core/src/Box/Box.stories.tsx
@@ -68,6 +68,10 @@ export default {
       control: { type: "select" },
       options: allColors,
     },
+    border: {
+      control: { type: "radio" },
+      options: ["primary", "none"],
+    },
     dangerouslySetInlineStyle: {
       control: { type: "object" },
     },
@@ -171,7 +175,7 @@ export default {
       control: "text",
     },
     rounding: {
-      options: ["none", "sm", "md", "lg", "xl", "full"],
+      options: ["none", "sm", "md", "full"],
       control: { type: "select" },
     },
     children: {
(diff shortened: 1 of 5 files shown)


btorndorff (approved):
LGTM

packages/syntax-core/src/Box/Box.tsx
```diff
@@ -98,6 +99,15 @@ type BoxProps = {
    * The background color of the box.
    */
   backgroundColor?: (typeof allColors)[number];
+  /**
+   * The border styling of the box.
+   *
+   * * `primary`: 1px solid var(--color-base-primary)
```
christianvuerings, line 105: @somethiiing could we update this to the actual value?

```suggestion
   * * `primary`: 1px solid var(--color-cambio-gray-370)
```

packages/syntax-core/src/Box/Box.stories.tsx
```diff
@@ -406,7 +420,7 @@ const roundingLookup = {
 export const Rounding: StoryObj<typeof Box> = {
   render: () => (
     <Box display="flex" gap={4} flexWrap="wrap">
-      {(["sm", "md", "lg", "xl", "full"] as const).map((rounding) => (
+      {(["sm", "md", "full"] as const).map((rounding) => (
```
christianvuerings, line 423: @somethiiing Thanks for catching these!

PR #386, opened 2024-04-15
Colors/Box: remove cambio naming + add all colors to Box Background color

somethiiing:
Some of the old colors were clashing with the new cambio colors (things like gray200 were not matching) i've updated them now, and have removed the old deprecated colors.

[figma](https://www.figma.com/file/G3UM2urgAYO2iiNU7nlulI/%F0%9F%9F%A1-Cambio-Syntax?type=design&node-id=0-1&mode=design&t=iYKoGMK1zNXPytG6-0)

This might not actually be a breaking change, I only changed the internals of `Color`, but it was mostly just in case. 

Also mostly updated broken stories, etc


btorndorff (approved):
LG

.changeset/itchy-kids-brush.md
```diff
@@ -0,0 +1,9 @@
+---
+"@cambly/syntax-core": major
+---
+
+Box: Remove depecated colors + update naming
```
nathan-ong, line 5: nit: typo "deprecated"

nathan-ong (approved):

zprong:
✅

PR #391, opened 2024-04-16
Colors: re-add missing primary colors

somethiiing:


packages/syntax-core/src/colors/colors.module.css
```diff
@@ -168,7 +168,31 @@
   box-shadow: 0 0 0 1px var(--color-cambio-destructive-370) inset;
 }
 
-/* Border Colors */
+/* Background Colors */
+.primary100BackgroundColor {
```
zprong, line 172: would this have been caught in CI if we didn't change [this line](https://github.com/Cambly/syntax/pull/386/files#diff-17d176aefcb5c3432a5b8bfd714d854c89dd192be8bdc1ed1d31a73ab98a848fL330)?

zprong (approved):

packages/syntax-core/src/colors/colors.module.css
somethiiing: the background color? or the comment. the comment no. the missing background color breaks cambly-frontend (typescript catches it)

zprong: Could we have kept the `backgroundColor="primary700"` there? TS would catch it there right?

somethiiing: i followed the [figma](https://www.figma.com/file/G3UM2urgAYO2iiNU7nlulI/%F0%9F%9F%A1-Cambio-Syntax?type=design&node-id=698-13800&mode=design&t=UnGaQocfWTUO1FgS-0) and it didnt have the primary colors, so for some reason i thought we changed how we were naming stuff. so its really my bad

zprong: ahh got it

PR #400, opened 2024-05-07
TextArea/add-ability-to-not-resize

somethiiing:
Add optional resize prop to allow for disabling resize.

image

image

This prop was needed for ai assessments to not allow for text area resizing

### DIFF
--- a/packages/syntax-core/src/TextArea/TextArea.tsx
+++ b/packages/syntax-core/src/TextArea/TextArea.tsx
@@ -44,6 +44,16 @@ type TextAreaProps = {
    */
   placeholder?: string;
   /**
+   * Sets which part resizes when the user drags the resize handle.
+   * * `none`: TextArea can not be resized
+   * * `horizontal`: TextArea can only be resized horizontally
+   * * `vertical`: TextArea can only be resized vertically
+   * * `both`: TextArea can be resized horizontally and vertically
+   *
+   * @defaultvalue `none`
+   */
+  resize?: "none" | "horizontal" | "vertical" | "both";
+  /**
    * Number of rows to display
    */
   rows?: number;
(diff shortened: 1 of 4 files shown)


packages/syntax-core/src/TextArea/TextArea.tsx
```diff
@@ -51,6 +51,12 @@ type TextAreaProps = {
    * Value of the TextArea
    */
   value: string;
+  /**
+   * Boolean of whether the TextArea should be resizable
+   *
+   * @default true
+   */
+  resize?: boolean;
```
christianvuerings, line 59: @somethiiing could we change this to take in the values of the CSS resize attribute? That way we avoid the boolean attribute and allow developers to be more specific:
https://developer.mozilla.org/en-US/docs/Web/CSS/resize

```suggestion
  resize?: "none" | "horizontal" | "vertical" | "both";
```

packages/syntax-core/src/TextArea/TextArea.tsx
```diff
@@ -51,6 +51,12 @@ type TextAreaProps = {
    * Value of the TextArea
    */
   value: string;
+  /**
+   * Boolean of whether the TextArea should be resizable
+   *
+   * @default true
```
christianvuerings, line 57: @somethiiing should we make the default `none`? I think the current resizing behavior (default "on") breaks in a lot of cases

packages/syntax-core/src/TextArea/TextArea.tsx
christianvuerings: @somethiiing one day we can also add something like `auto` to automatically resize the text area once browser support is here: https://chriscoyier.net/2023/09/29/css-solves-auto-expanding-textareas-probably-eventually/

packages/syntax-core/src/TextArea/TextArea.tsx
somethiiing: oh, i thought the default should have been none. but i was more concerned about backwards compatability. i can change it

packages/syntax-core/src/TextArea/TextArea.tsx
somethiiing: awesome! ill keep a an eye out for when enough browser support happens!

somethiiing: also, updated with css resize attribute types!

christianvuerings (approved):
@somethiiing thanks for the changes - please address comments before merging

PR #411, opened 2024-05-10
RadioButton/Checkbox: Add describedby as a prop

somethiiing:
for https://axeauditor.dequecloud.com/test-run/de95ac14-f88c-11ee-a30d-2b2389e5f0d8/issue/07ac1b62-fc82-11ee-a631-135645ea1d04?sortField=ordinal&sortDir=asc&page=0&issuesCount=&searchKeyword=1724726&searchOn=id&row=0

the fix that deque recommends are:
image

I already recommended to @jennyhe1120 that she already do the second recommendation (add the fieldname to the error message), but since deque _strongly_ recommends the first option, i created this POC.

I created an ariaProps object that can be reused between the different inputs (checkbox/radio/textinput/etc), i only put `aria-describedby` for now, because that's the only one we need. other suggested ones are: `aria-label`, `aria-labelledby`, `aria-hidden` but concluded we didnt need them yet.

the other option was the just manually map the prop `aria-describedby` onto the the component, but that felt unwieldy if we ever added more (like in some places we already use `aria-label`)

### DIFF
--- a/packages/syntax-core/src/Checkbox/Checkbox.stories.tsx
+++ b/packages/syntax-core/src/Checkbox/Checkbox.stories.tsx
@@ -2,6 +2,7 @@ import { type StoryObj, type Meta } from "@storybook/react";
 import Checkbox from "./Checkbox";
 import React, { useState } from "react";
 import Box from "../Box/Box";
+import Typography from "../Typography/Typography";
 
 export default {
   title: "Components/Checkbox",
@@ -19,6 +20,7 @@ export default {
     error: false,
     size: "md",
     "data-testid": "",
+    "aria-describedby": "",
   },
   argTypes: {
     checked: {
@@ -42,8 +44,10 @@ export const Default: StoryObj<typeof Checkbox> = {};
 
 const CheckboxInteractive = ({
   label = "checkbox label",
+  "aria-describedby": ariaDescribedby,
 }: {
   label?: string;
+  "aria-describedby"?: string;
 }) => {
   const [isChecked, setIsChecked] = useState(false);
   const handleChange = () => {
@@ -51,7 +55,7 @@ const CheckboxInteractive = ({
   };
   return (
     <Checkbox
-      ariaValues={{ "aria-describedby": "checkboxes" }}
+      aria-describedby={ariaDescribedby}
       checked={isChecked}
       onChange={handleChange}
       label={label}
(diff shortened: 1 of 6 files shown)


somethiiing:
@christianvuerings can you give input on how you'd handle the issue described in the pr description? if this POC looks okay, i can add tests/update the remaining form input components

packages/syntax-core/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -49,7 +49,14 @@ const CheckboxInteractive = ({
   const handleChange = () => {
     setIsChecked(!isChecked);
   };
-  return <Checkbox checked={isChecked} onChange={handleChange} label={label} />;
+  return (
+    <Checkbox
+      ariaValues={{ "aria-describedby": "checkboxes" }}
```
christianvuerings, line 54: @somethiiing does the `checkboxes` id actually exist on this page? If not, could we add it so we can test the behavior?

packages/syntax-core/src/Checkbox/Checkbox.tsx
```diff
@@ -68,6 +70,12 @@ const Checkbox = ({
    * @defaultValue false
    */
   error?: boolean;
+  /**
+   * Whether or not there is an error with the input
```
christianvuerings, line 74: @somethiiing the description doesn't make a lot of sense - can we make it more specific?

packages/syntax-core/src/Checkbox/Checkbox.tsx
```diff
@@ -68,6 +70,12 @@ const Checkbox = ({
...
+   *
+   * @defaultValue false
+   */
+  ariaValues?: ariaProps;
```
christianvuerings, line 78: @somethiiing I would avoid passing in an object of props when possible. Just use `aria-describedby` instead.

christianvuerings: Example as to why we want to avoid it:
image

christianvuerings (approved):
@somethiiing going to approve to unblock but please make sure to address comments before merging.

- [x] Add `checkboxes` (or other element which uses aria describedby in the story book)
- [x] Update docs
- [x] Flatten props
- [x] Add `aria-describedby` to `RadioButton` as well

packages/syntax-core/src/Checkbox/Checkbox.tsx
somethiiing: oops, i had copied and pasted the above error to get the same formatting, but forgot to change the actual error message

PR #419, opened 2024-05-15
Icons/generate script

somethiiing:
major files are just the generate-icon script + template
uploaded new svg files and ran the script to generate new ones.
docs: https://react-svgr.com/docs/custom-templates/

Syntax/Icons:
- add icon generation script
- update icons per design
- add 2 new icons


packages/syntax-icons/src/Icons.stories.tsx
```diff
@@ -39,7 +39,7 @@ import HeartFilled from "./icons/HeartFilled";
 import HeartUnfilled from "./icons/HeartUnfilled";
 import Help from "./icons/Help";
 import Home from "./icons/Home";
-import IdentifyDocument from "./icons/IdentifyDocument";
+import IdentifyDocument from "./icons/IdentityDocument";
```
somethiiing, line 42: @christianvuerings  you mentioned before that i should add imports with the script using AST, do we have an example somewhere i can see how to do that?

somethiiing: also, can i update the package-json exports the same way?

package.json
```diff
@@ -13,7 +13,8 @@
     "test:watch": "turbo run test:watch",
     "test": "turbo run test",
     "version-packages": "changeset version && pnpm install --no-frozen-lockfile",
-    "generate": "npx tsx './generate-component/' $1"
+    "generate": "npx tsx './generate-component/' $1",
+    "icons": "npx tsx './packages/syntax-icons/src/generate-icons/generateIcons.ts'"
```
christianvuerings, line 17:
```suggestion
    "generate-icons": "npx tsx ./packages/syntax-icons/src/generate-icons/generateIcons.ts"
```

packages/syntax-icons/src/generate-icons/generateIcons.ts
```diff
@@ -0,0 +1,80 @@
...
+    )
+    .replace(/^[a-z]/, (group) => group.toUpperCase());
+}
+
+function getFileName(file: string) {
+  return snakeToPascal(file.replace(".svg", ""));
+}
+
+async function generateIcons() {
+  try {
```
christianvuerings, line 32: @somethiiing do we even need the `try` / `catch` if we just do `throw error` in the `catch`?

packages/syntax-icons/src/generate-icons/generateIcons.ts
```diff
@@ -0,0 +1,80 @@
...
+        jsCode,
+      );
+    }
+  } catch (error) {
+    throw error;
+  }
+}
+
+function runPrettier(): void {
+  exec("pnpm format", (error) => {
```
christianvuerings, line 64: @somethiiing would be good to only run prettier on the icon files. We can return those paths from `generateIcons`

packages/syntax-icons/src/generate-icons/iconTemplate.tsx
```diff
@@ -0,0 +1,20 @@
...
+const iconTemplate: Template = ({ componentName, jsx }, { tpl }) => {
+  return tpl`
+  import { type ComponentProps, forwardRef } from "react";
+  import Icon from "../../../syntax-core/src/Icon/Icon";
+
+  const ${componentName} = forwardRef<
+    SVGSVGElement,
+    Omit<ComponentProps<typeof Icon>, "path">
+  >(({ color, size }, ref) => (
+    <Icon ref={ref} path="${jsx.children[0].openingElement.attributes[0].value.value}" color={color} size={size} />
```
christianvuerings, line 12: @somethiiing this seems a bit dangerous - is there any other way to get the `path`?

packages/syntax-icons/src/generate-icons/iconTemplate.tsx
```diff
@@ -0,0 +1,20 @@
...
+  ));
+
+  ${componentName}.displayName = "${componentName}";
+  export default ${componentName}
+  `;
+};
+
+export default iconTemplate;
```
christianvuerings, line 20: @somethiiing would be great if we can use `export default function iconTemplate` instead of using `const` to define the function.

packages/syntax-icons/package.json
```diff
@@ -387,5 +388,9 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "dependencies": {
```
christianvuerings, line 392: @somethiiing super important - these should be `devDependencies`, not dependencies! Otherwise this will be included in the npm install of `@cambly/syntax-icons`

packages/syntax-icons/src/Icons.stories.tsx
christianvuerings: @somethiiing I would suggest doing that in a follow up and using our code mods to do that: https://github.com/Cambly/syntax/tree/main/packages/syntax-codemods/src/transforms

Another approach would be to just write a specific ESLint rule for both files (imports + package.json) which would auto fix them.
For supporting `.json` with ESLint you can use something like https://github.com/ota-meshi/eslint-plugin-jsonc

packages/syntax-icons/src/generate-icons/generateIcons.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (error) {
+      // eslint-disable-next-line no-console
+      console.error(`Error: ${error.message}`);
```
christianvuerings, line 67: @somethiiing can we `throw error` instead? We do want to be notified if we run into a prettier error

christianvuerings (approved):
@somethiiing looking great, this will definitely improve maintainability of our icons. Please do address comments before merging

packages/syntax-icons/src/generate-icons/iconTemplate.tsx
somethiiing: i forgot to tag you in this part, this was specifically the one where i didnt know how to solve/resolve. wasnt sure if there was a better way to access this

packages/syntax-icons/src/generate-icons/iconTemplate.tsx
somethiiing: hm, this looks a little harder to read compared to the first variation:

image

i'm going to leave it as the original unless you feel more strongly about it

packages/syntax-icons/src/generate-icons/iconTemplate.tsx
somethiiing: this is a kind of ridiculous looking if statement, but it solves the typing error:

image

but because they use AST, (and knowing my knowledge from building the linter), this does feel like the safest/correct-est way to fix it

somethiiing: @christianvuerings just leaving this for the night, if you dont respond by the morning, ill merge it as is

PR #423, opened 2024-05-19
Typography/Heading: update sizes

christianvuerings:
# Description

Update sizes according to https://www.figma.com/design/G3UM2urgAYO2iiNU7nlulI/%F0%9F%9F%A1-Cambio-Syntax?node-id=601-5179&m=dev

# Changes

* Add size `000`
* Remove sizes: `500`, `600`, `800` and `1000`
* Adjust size values

# Open questions

## We can’t call the size `000` on web, can we rename it to `0` instead?

@ajmooo ?

Screenshot 2024-05-19 at 6 29 32 AM

## What should the removed sizes map to?

@ajmooo  ?

| Before | After |
| ---- | ---- |
| `500` | ? |
| `600` | ? |
| `800` | ? |
| `1000` | ? |

## The default size for heading used to be 500, what should it be now?

@ajmooo ?


somethiiing (approved):

ajmooo (commented):
lgtm

PR #430, opened 2024-05-22
Tooltip: add on prop to adapt dark or white background

ASkyBig:
By default, on="lightBackground'. And if you want to use tooltip on a dark background, add on="darkBackground".

image

### DIFF
--- a/.changeset/chatty-boxes-judge.md
+++ b/.changeset/chatty-boxes-judge.md
@@ -1,0 +1,6 @@
+---
+"@cambly/syntax-core": minor
+"@syntax/storybook": minor
+---
+
+add prop on for tooltip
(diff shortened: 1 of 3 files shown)


.changeset/chatty-boxes-judge.md
```diff
@@ -0,0 +1,6 @@
+---
+"@cambly/syntax-core": major
```
jliotta, line 2: @somethiiing should this just be a minor change?

somethiiing: yeah, major changes are for things that are breaking. such as removing a prop, or changing a propname, etc. add-ons or not breaking changes should all be minor

ASkyBig: got it!

jliotta (approved):

PR #431, opened 2024-05-22
Syntax: increase memory + fix build

somethiiing:
increase memory size for build script

similar solution to our frontend knip problem: https://github.com/Cambly/Cambly-Frontend/blob/bcf887d3d3f6538fac0d5328923c833693e55e88/.github/workflows/knip.yml#L30

image


jliotta (approved):

zprong:
nice B)

PR #433, opened 2024-05-22
TextField: Add step param

williamnewman94:
Need step for this: https://www.figma.com/design/FDcPtWKnUQWY8VjVFhpjxe/Availability-filter?node-id=10699-17858&m=dev. So I added it.

Official documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step

I also fixed the other stories so they actually work:
Before:
Screenshot 2024-05-22 at 5 33 06 PM

After:
Screenshot 2024-05-22 at 5 32 38 PM

### DIFF
--- a/packages/syntax-core/src/TextField/TextField.stories.tsx
+++ b/packages/syntax-core/src/TextField/TextField.stories.tsx
@@ -21,7 +21,7 @@ export default {
     "data-testid": "",
     id: "",
     value: "",
-    step: 0,
+    step: 1,
   },
   argTypes: {
     autoComplete: {
--- a/packages/syntax-core/src/TextField/TextField.stories.tsx
+++ b/packages/syntax-core/src/TextField/TextField.stories.tsx
@@ -21,6 +21,7 @@ export default {
     "data-testid": "",
     id: "",
     value: "",
+    step: 0,
   },
   argTypes: {
     autoComplete: {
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/TextField/TextField.stories.tsx
```diff
@@ -89,13 +90,21 @@ export const Default: StoryObj<typeof TextField> = {
...
 export const helperText: StoryObj<typeof TextField> = {
-  render: (args) => <TextFieldDefault helperText="Helper text" {...args} />,
+  args: { helperText: "Helper text" },
+  render: (args) => <TextFieldDefault {...args} />,
 };
 
 export const ErrorText: StoryObj<typeof TextField> = {
-  render: (args) => <TextFieldDefault errorText="This is an error" {...args} />,
+  args: { errorText: "This is an error" },
+  render: (args) => <TextFieldDefault {...args} />,
```
williamnewman94, line 99: These were all broken, see in the description for before and after.

packages/syntax-core/src/TextField/TextField.tsx
```diff
@@ -80,6 +81,10 @@ export default function TextField({
    * Value of the TextField
    */
   value: string;
+  /**
+   * Specified legal number intervals for an input field. Specifically for time or number. If for time specify for milliseconds. Must be a positive value.
```
williamnewman94, line 85: It needs to be a positive number but there is no clear way to enforce that in TS sadly so just point it out in the docs.
For reference: https://stackoverflow.com/questions/21224922/is-there-a-way-to-represent-a-non-negative-integer-in-typescript-so-that-the-com.

zprong, line 85: nit: a bit hard to read
```suggestion
   * Specified legal number intervals for an input field. Specifically for time or number. If for time, specify in milliseconds. Must be a positive value.
```

packages/syntax-core/src/TextField/TextField.stories.tsx
```diff
@@ -21,6 +21,7 @@ export default {
     "data-testid": "",
     id: "",
     value: "",
+    step: 0,
```
zprong, line 24: This should be 1 right?

zprong (commented):
LG pending comments

zprong (approved):

PR #427, opened 2024-05-21
Tabs: add component

zprong:
Figma:
https://www.figma.com/design/G3UM2urgAYO2iiNU7nlulI/%F0%9F%9F%A1-Cambio-Syntax?node-id=4540-1486&t=GyzLoUF9f8Sq4Pli-0

Pending:
- Waiting for [this ticket](https://cambly.atlassian.net/browse/SYN-62?atlOrigin=eyJpIjoiZDZkYmUyZjUyNDI0NGEyZjgzMTZmMjJmMGY2MDA4MjUiLCJwIjoiaiJ9) to be done to allow for `white-70` (text color on dark background) and `white-40` on the container's bottom border

Open questions [ANSWERED [BELOW](https://github.com/Cambly/syntax/pull/427#issuecomment-2123451151)]
- Typography is supposed to be 000-semibold-caps, but that's not supported (going to check in with AJ tomorrow)
- We don't have a white-secondary font, but Figma specifies it for unselected tabs on dark background
- Should dark background have the spanned underline across the tabs container?

Things to ignore:
- Design specifies a full-width bottom-border on the Tab bar if it changes URLs - AJ said we can skip this requirement for now if it adds too much complexity (but maybe it's not that hard?)

Tab Buttons
image

Tab Links
image

PR #440, opened 2024-06-04
Tabs: default button and link to be on light background

zprong:
## Changes
- defaults tab buttons and tab links to default to light background
- upgrades version for syntax-icons to update the expected color prop type for icons
- cleans up our TypeScript errors that currently exist in the repo

## The immediate problem
We can't upgrade Syntax to 15.4.0 ([PR](https://github.com/Cambly/syntax/pull/427)) and I believe it's actually because there should've also been an upgrade to `syntax-icons`

FE Upgrade attempt for reference: https://github.com/Cambly/Cambly-Frontend/pull/10243

Here we can see
```
Types of parameters 'props' and 'props' are incompatible.
      Type 'IconProps & RefAttributes<SVGSVGElement>' is not assignable to type 'Omit<Omit<{ color?: "white" | "success" | "inherit" | "primary" | "gray700" | "gray900" | "destructive-primary" | "destructive-darkBackground" | "success-darkBackground" | undefined; size?: "sm" | ... 3 more ... | undefined; path?: string | undefined; } & RefAttributes<...>, "path">, "ref"> & RefAttributes<...>'.
Types of property 'color' are incompatible.
Type '"white" | "success" | "inherit" | "primary" | "gray700" | "gray900" | "destructive-primary" | "destructive-darkBackground" | "success-darkBackground" | "white-secondary" | undefined' is not assignable to type '"white" | "success" | "inherit" | "primary" | "gray700" | "gray900" | "destructive-primary" | "destructive-darkBackground" | "success-darkBackground" | undefined'.
```

Here, `Omit<Omit<{ color?:...}` does not contain `color: white-secondary` but below, you can see that it's properly typed
image

## The long-term problem
In that case, we'd run into this issue with our Syntax Icons + IconButton anytime we introduce a new color to Typography (Icon color is based on Typography color).


somethiiing:
@christianvuerings can you give us access to the vercel app? i've ran into this issue before but not having access and trying to blindly figure out what's wrong with the deploy is kind of sandbagging development

somethiiing (approved):

PR #441, opened 2024-06-04
Icon: Update icon sizing

jliotta:
Per a request from Joey and AJ, we're updating `Icon` sizing to be more like `Typography`. Will run a codemod in Cambly-FE after this is merged to update the sizing of our syntax icons

### DIFF
--- a/packages/syntax-core/src/Chip/Chip.tsx
+++ b/packages/syntax-core/src/Chip/Chip.tsx
@@ -123,7 +123,7 @@ const Chip = forwardRef<HTMLButtonElement, ChipProps>(
         aria-pressed={selected}
         onClick={onChange}
       >
-        {Icon && <Icon className={iconStyles} size="sm" />}
+        {Icon && <Icon className={iconStyles} size={100} />}
         <Box paddingX={Icon ? 1 : 0}>
           <Typography size={100} color={color} weight="medium">
             {text}
--- a/packages/syntax-core/src/Icon/Icon.tsx
+++ b/packages/syntax-core/src/Icon/Icon.tsx
@@ -14,14 +14,20 @@ type IconProps = {
   color?: ComponentProps<typeof Typography>["color"];
   /**
    * The size of the Icon.
-   * * `xs`: 16px x 16px
-   * * `sm`: 20px x 20px
-   * * `md`: 24px x 24px
-   * * `lg`: 32px x 32px
+   * * 100: 16px x 16px
+   * * 200: 20px x 20px
+   * * 300: 24px x 24px
+   * * 400: 32px x 32px
+   * * 500: 48px x 48px
+   * * 600: 72px x 72px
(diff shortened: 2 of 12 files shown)


packages/syntax-core/src/Chip/Chip.tsx
```diff
@@ -123,7 +123,7 @@ const Chip = forwardRef<HTMLButtonElement, ChipProps>(
         aria-pressed={selected}
         onClick={onChange}
       >
-        {Icon && <Icon className={iconStyles} size="sm" />}
+        {Icon && <Icon className={iconStyles} size={100} />}
         <Box paddingX={Icon ? 1 : 0}>
```
jliotta, line 127: @christianvuerings @somethiiing The issue I'm getting with these other components that use `Icon` like `Chip`, `IconButton` etc is that we are still using both Material icons and our own Icon.

Now with this change, the sizing of them is different. Material uses `sm` `med` `lg` and our Icon component will now use `100` `200` etc. 

How do I determine what type of Icon is being passed in as a prop?

somethiiing: cant you just pass in both props?

jliotta: They both use the name `size` as the prop.

I think I got it to work though. The `size` prop that's passed down controls the size for our internal icons and for material the className that was being used controls the size of the icon

packages/syntax-core/src/Icon/Icon.tsx
```diff
@@ -14,14 +14,20 @@ type IconProps = {
   color?: ComponentProps<typeof Typography>["color"];
   /**
    * The size of the Icon.
-   * * `xs`: 16px x 16px
-   * * `sm`: 20px x 20px
-   * * `md`: 24px x 24px
-   * * `lg`: 32px x 32px
+   * * 100: 16px x 16px
```
somethiiing, line 17: missing the size 0?

somethiiing (approved):

packages/syntax-core/src/Icon/Icon.tsx
jliotta: Joey didn't spec out a size 0. we can easily add it in the future if that changes

packages/syntax-core/src/IconButton/IconButton.tsx
```diff
@@ -8,12 +8,18 @@ import { backgroundColor } from "../colors/backgroundColor";
...
 import type InternalIcon from "../Icon/Icon";
 
-const iconSize = {
+const materialIconSize = {
   sm: styles.smIcon,
   md: styles.mdIcon,
   lg: styles.lgIcon,
 };
 
+const internalIconSize = {
```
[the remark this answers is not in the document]
jliotta: yeah thanks for the catch

PR #443, opened 2024-06-07
Icon: Update sizing to match Typography scale

jliotta:
Small follow-up to the previous update. Noticed that the Icon scale as speced didnt exactly match the one with Typography (no size 0, default 300 instead of 200). Confirmed with design that we should update it to this.

### DIFF
--- a/packages/syntax-core/src/Icon/Icon.tsx
+++ b/packages/syntax-core/src/Icon/Icon.tsx
@@ -14,16 +14,16 @@ type IconProps = {
   color?: ComponentProps<typeof Typography>["color"];
   /**
    * The size of the Icon.
-   * * 100: 16px x 16px
-   * * 200: 20px x 20px
-   * * 300: 24px x 24px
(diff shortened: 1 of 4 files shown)


packages/syntax-core/src/Icon/Icon.tsx
```diff
@@ -14,16 +14,16 @@ type IconProps = {
...
-   * * 500: 48px x 48px
-   * * 600: 72px x 72px
-   * * 700: 100px x 100px
-   * * 800: 140px x 140px
-   * * 900: 200px x 200px
-   * * 1000: 280px x 280px
+   * * 0: 16px x 16px
+   * * 100: 20px x 20px
+   * * 200: 24px x 24px
+   * * 300: 28px x 28px
```
jliotta, line 20: AJ wanted to adjust the large size down to fit better in icon buttons

somethiiing (approved):

PR #446, opened 2024-06-09
Tabs.Link: fix selected state on=darkBackground

christianvuerings:
# Background

Noticed the following issue in VSCode:

> Property 'selectedTabDarkBackground' does not exist on type '{ tabContainerlightBackground: string; tabContainerDarkBackground: string; selectedTabOnLightBackground: string; selectedTabOnDarkBackground: string; unselectedTab: string; link: string; }'. Did you mean 'selectedTabOnDarkBackground'?

Screenshot 2024-06-09 at 8 11 14 AM

Note that this TS issue doesn't fail the build because of this outstanding TS ticket to support plugins during compilation:
* https://www.npmjs.com/package/typescript-plugin-css-modules#about-this-plugin
* https://github.com/microsoft/TypeScript/issues/16607

# Before (incorrect selected state)
Screenshot 2024-06-09 at 8 16 50 AM

# After
Screenshot 2024-06-09 at 8 17 05 AM

# Notes

Also added a light / dark toggle to the `Tabs.Link` example so we can easily test: 
Screenshot 2024-06-09 at 8 26 29 AM

### DIFF
--- a/packages/syntax-core/src/Tabs/TabLink.tsx
+++ b/packages/syntax-core/src/Tabs/TabLink.tsx
@@ -57,7 +57,7 @@ const TabLink = forwardRef<HTMLAnchorElement, TabLinkProps>(
           [styles.unselectedTab]: !selected,
           [styles.selectedTabOnLightBackground]:
             selected && on === "lightBackground",
-          [styles.selectedTabDarkBackground]:
+          [styles.selectedTabOnDarkBackground]:
             selected && on === "darkBackground",
         })}
         style={{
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/Tabs/TabLink.tsx
```diff
@@ -57,7 +57,7 @@ const TabLink = forwardRef<HTMLAnchorElement, TabLinkProps>(
           [styles.unselectedTab]: !selected,
           [styles.selectedTabOnLightBackground]:
             selected && on === "lightBackground",
-          [styles.selectedTabDarkBackground]:
```
zprong, line 60: 🤦

zprong (approved):
Thank you for catching + fixing this!

PR #450, opened 2024-06-11
Card: Add ref so impressions can be used on it

somethiiing:


jliotta (approved):

packages/syntax-core/src/Card/Card.tsx
```diff
@@ -28,22 +29,36 @@ type CardType = {
 
 /**
  * [Card](https://cambly-syntax.vercel.app/?path=/docs/components-card--docs) is a basic container component to apply consistent styling and render child components.
+ * const Box = forwardRef<HTMLDivElement, BoxProps>(function Box(
+  props: BoxProps,
+  ref,
+): ReactElement {
+  const { as: BoxElement = "div", children, ...boxProps } = props;
```
zprong, line 36: is this supposed to be here?

PR #455, opened 2024-06-24
Color: Gray Value Updates

somethiiing:
Hex values for gray700/800/870/900 have been updated:

gray700 → #5E5952

gray800 → #363432

gray870 → rgba(94, 89, 82, 0.7)

gray900 → #262625

 

Add 

gray270 → rgba(228, 219, 211, 0.7)

white70 → rgba(255, 255, 255, 0.7)

white40 → rgba(255, 255, 255, 0.4)


.github/workflows/release.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+permissions:
```
zprong, line 8: What does this line do?

.changeset/slow-carrots-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cambly/syntax-icons": major
```
zprong, line 2: Why does this need to be major?

btorndorff (approved):

.github/workflows/release.yml
somethiiing: was part of an old commit

.changeset/slow-carrots-shop.md
somethiiing: was part of an old commit

PR #461, opened 2024-06-27
DesignTokens: add shadows to design token

somethiiing:
add shadow to design token

https://www.figma.com/design/G3UM2urgAYO2iiNU7nlulI/Cambio-Syntax?node-id=6478-334&t=eaf3iqO614IAJAyz-0


zprong (approved):
Just so I have this right: we won't be able to use this in Cambly-Frontend until this gets plugged in somewhere in syntax right?

PR #481, opened 2024-07-16
selectlist: update arrow icon color

somethiiing:


Farzan-Hashmi (approved):
lgtm

packages/syntax-core/src/SelectList/SelectList.tsx
```diff
@@ -111,6 +112,18 @@ export default function SelectList({
     clear: "white",
   };
 
+  const getArrowIconColor = () => {
+    if (errorText) {
+      return ColorBaseDestructive700;
+    } else {
```
nathan-ong, line 118: nit: should this just be 

```
if (errorText) {
  return ColorBaseDestructive700
} else if (color === "clear") {
  return ColorCambioWhite100;
} else { 
  return ColorBaseGray700;
}
```

somethiiing: oops, yours is a little more clear, but i dont think its a big deal to leave it this way. ill tweak it next time im in theses files

PR #498, opened 2024-07-26
Tabs: update underline css

zprong:
Context: https://cambly.slack.com/archives/C033ZPY5M46/p1720808650914279

The underline on the selected tab's margin is slightly off from design

Current:
image

Ideal:
image


btorndorff (approved):
LG

PR #510, opened 2024-07-31
Avatar: update icon sizes and offsets for medium and small variants

zprong:
I got the following specs wrong on the first iteration:

medium avatars should have a 16x16 icon with a lower offset 
small avatars should have a 12x12 icon instead of 10x10


btorndorff (approved):
LG

PR #512, opened 2024-08-01
ButtonGroup: Make orientation responsive

zprong:
Currently there are a few usages of ButtonGroup where the CTAs could be more responsive. Things will look normal on a desktop but then the mWeb view looks squished because CTAs are still side-by-side. Allowing ButtonGroup's orientation to be responsive should help alleviate most of those issues.

New props:
`smOrientation`
`lgOrientation`

### DIFF
--- a/.changeset/clever-balloons-change.md
+++ b/.changeset/clever-balloons-change.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+ButtonGroup: Make responsive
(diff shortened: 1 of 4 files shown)


btorndorff (commented):
Can you attach a video of what this looks like now? I am wondering if flex-wrap is a solution to this but otherwise seems fine.

zprong:
sure thing! I think flex-wrap could work too

https://github.com/user-attachments/assets/07416cf8-4da0-4450-b294-39f36a7b0ed1

somethiiing:
> Can you attach a video of what this looks like now? I am wondering if flex-wrap is a solution to this but otherwise seems fine.

im not _entirely_ sure, but if there were (more than 2 buttons), it would pop them down one at a time. whereas this one would change the orientation in a more predicatable way

somethiiing (approved):

PR #519, opened 2024-08-07
WordConfetti: Fix WordConfetto key

jliotta:
Noticed an error in the console when developing in Cambly-FE about the key in WordConfetto. Removed it as an actual prop and fixed the issue. Tested by looking at the console in the local storybook

Screenshot 2024-08-06 at 5 16 04 PM


williamnewman94 (approved):
LG!

PR #544, opened 2024-09-03
Only publish ESM version of packages & validate type exports

christianvuerings:
# Description

Since we now use `"module": "type"` in Cambly-Frontend, we can resolve to only export ESM modules. We should also validate the types that we're exporting.

# Changes

* Only publish ESM version of packages
* Validate type exports with [Are the types wrong](https://arethetypeswrong.github.io/)

# Reviewer notes

* `check-exports` relies on the build to be completed
* The `pack` command will package the package into a `.tgz`. We need to make sure those files are not committed to the repo
* Ignored `cjs-resolves-to-esm` since we only publish esm - also see https://www.totaltypescript.com/how-to-create-an-npm-package#56-fix-the-cjs-warning
* Ignored `no-resolution` for syntax-icons. See https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/d129c2b898619614486e63cf40d968c9571e88f1/docs/problems/NoResolution.md#true-positive-node-10-doesnt-support-packagejson-exports & #417 -


christianvuerings:
@williamnewman94 just fixed the merge conflict. Would be great if you take a look at this sometime today

packages/syntax-icons/package.json
```diff
@@ -4,18 +4,18 @@
...
   ],
   "scripts": {
-    "build": "tsup src/icons/*.tsx --format esm,cjs --dts --external react",
-    "dev": "tsup src/icons/*.tsx --format esm,cjs --watch --dts --external react",
+    "build": "tsup src/icons/*.tsx --format esm --dts --external react",
+    "check-exports": "attw $(pnpm pack) --ignore-rules=cjs-resolves-to-esm --ignore-rules=no-resolution",
+    "dev": "tsup src/icons/*.tsx --format esm --watch --dts --external react",
     "lint": "TIMING=1 eslint \"src/**/*.ts*\" --max-warnings 0",
-    "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist",
+    "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist && rm -rf *.tgz",
```
williamnewman94, line 18: We're we exporting `tgz` when we didn't need to?

packages/syntax-icons/package.json
```diff
@@ -4,18 +4,18 @@
...
   "types": "./dist/index.d.ts",
   "sideEffects": false,
   "license": "Apache-2.0",
   "files": [
     "dist/**"
   ],
   "scripts": {
-    "build": "tsup src/icons/*.tsx --format esm,cjs --dts --external react",
-    "dev": "tsup src/icons/*.tsx --format esm,cjs --watch --dts --external react",
+    "build": "tsup src/icons/*.tsx --format esm --dts --external react",
```
williamnewman94, line 14: The same as before just omitted cjs?

tsup.config.ts
```diff
@@ -17,7 +17,7 @@ export default defineConfig({
       "**/*.test.*",
     ],
   }),
-  format: ["cjs", "esm"],
+  format: "esm",
```
williamnewman94, line 20: Why did you use tsup instead of webpack? I chat-gpt'd it but curious what you say. Maybe it's obvious if you're building a components library you should use tsup.

packages/syntax-icons/package.json
williamnewman94: Going off the name does `check-exports` generate a tarball to check if the exports are esm-only then now we have to clean up the tarball? Why is this necessary?

turbo.json
```diff
@@ -5,6 +5,10 @@
       "outputs": ["dist/**", "storybook-static/**"],
       "dependsOn": ["^build"]
     },
+    "check-exports": {
+      "cache": false,
+      "dependsOn": ["^build"]
```
williamnewman94, line 10: This is basically saying wait for the `build` step to complete before running `check-exports`. I looked into turbo a bit, pretty cool. Was your intention to eventually use it in Cambly-Frontend? Was CI really slow for Syntax?

packages/syntax-icons/package.json
williamnewman94: Oh wait. `attw` is are the types wrong. It's generating the tarball then checking if the types are correct on it then deleting it. This makes sense but how is this different then typechecking? Is it more thorough? Or maybe its entirely different?

williamnewman94 (approved):
LG, some questions.

packages/syntax-icons/package.json
christianvuerings: @williamnewman94 see https://github.com/Cambly/syntax/pull/544/files#diff-3d7766fca98eabc07616b81b6551c3f305814a858011e90ff75263bb72d79159R15 - the `pack` command creates a tarball (`.tgz` file), so we do want to make sure it never gets committed to this repo & yes, that we do clean it up

packages/syntax-icons/package.json
christianvuerings: @williamnewman94 that's correct

tsup.config.ts
christianvuerings: @williamnewman94 tsup is more straightforward than webpack: less configuration setup & features to setup. It's also recommended for libraries like ours

turbo.json
christianvuerings: @williamnewman94 yes, that's correct. As for using `turbo` in `Cambly-Frontend`, I think that would only make sense if we move more into a monorepo setup with multiple packages (not something I immediately would recommend for us)

PR #554, opened 2024-09-23
Fix selectList darkbackground error on windows

ASkyBig:
There are some differences between select in Windows and Mac. On dark mode, if the background color is not set, it will cause the options container for select on Windows to have a default white background. Resulting in the inability to display white text. So we need to specify the background color for select
image

### DIFF
--- a/.changeset/empty-beers-listen.md
+++ b/.changeset/empty-beers-listen.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-core": minor
+---
+
+SelectList error on windows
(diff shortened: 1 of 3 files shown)


jliotta (approved):
I don't have a windows to test on so I'll trust you to test that. I did pull the changes locally and test on a mac and it still works as expected so I think it's safe to merge this.

I'm going to merge now and update the new package in Cambly-FE and deploy that. Once that's done you'll be able to test if the fix works on windows

PR #556, opened 2024-09-26
WordConfetti: add simple hashing function

jliotta:
With word confetti, we were still getting an issue of the word color and tilt changing on rerender (the useMemo originally implemented wasn't working).

This PR adds a simple hashing function so the same word will always have the same color and tilt

https://github.com/user-attachments/assets/2fd15dd7-3670-421c-8837-cd593a1330f4

### DIFF
--- a/packages/syntax-core/src/WordConfetti/WordConfetti.test.tsx
+++ b/packages/syntax-core/src/WordConfetti/WordConfetti.test.tsx
@@ -26,4 +26,22 @@ describe("wordConfetti", () => {
     expect(screen.getByText("test")).toBeInTheDocument();
     expect(screen.getByText("word")).toBeInTheDocument();
   });
+
+  it("returns a consistent color and tilt for words", () => {
+    render(
+      <WordConfetti
+        size={300}
+        theme="neutral"
+        words={["test", "word", "confetti"]}
+      />,
+    );
+
+    const confetti = screen.getByTestId("confetti");
+    expect(confetti.getAttribute("style")).toMatchInlineSnapshot(
+      `"padding: 16px 20px 16px 20px; transform: rotate(3deg);"`,
+    );
+    expect(confetti.className).toMatchInlineSnapshot(
+      `"_box_b2ac18 _thistleBackgroundColor_0d99d4"`,
+    );
+  });
 });
--- a/packages/syntax-core/src/WordConfetti/WordConfetti.tsx
+++ b/packages/syntax-core/src/WordConfetti/WordConfetti.tsx
@@ -1,4 +1,10 @@
-import { type ReactNode, forwardRef, useMemo, type ReactElement } from "react";
+import {
+  type ReactNode,
+  forwardRef,
+  useMemo,
+  type ReactElement,
+  useCallback,
+} from "react";
 import Box from "../Box/Box";
 import Typography from "../Typography/Typography";
 
@@ -8,7 +14,7 @@ const themeBackgroundColors = {
   warm: ["red", "tan", "orange"],
 } as const;
 
-const degreeOfTiltOptions = [-6, -6, -3, -3, 0, 3, 3, 6, 6];
+const degreeOfTiltOptions = [-6, -3, 0, 3, 6];
 
 const paddings = {
   300: "16px 20px 16px 20px",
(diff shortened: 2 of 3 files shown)


packages/syntax-core/src/WordConfetti/WordConfetti.tsx
```diff
@@ -104,15 +110,25 @@ const WordConfetti = forwardRef<HTMLDivElement, WordConfettiProps>(
       words,
     } = props;
 
+    const hashCode = useCallback((string: string) => {
```
christianvuerings, line 113: @jliotta should we re-write this to use `useMemo` instead? With `useCallback` we cache a function definition to ensure that the same function object is used across multiple renders. However with `useMemo` (what I think we want) we cache the return value of a function to avoid recalculating it on every render.

christianvuerings: @jliotta also, in this case, I would put the `hash` function outside of `WordConfetti`. Just makes it easier to ready through the code.

```tsx
function hash(input: string): number {
  //...
}

function WordConfetti() {
	const hashCode = useMemo(() => hash(word), [word]);
  return (
    //
  )
}
```

(didn't actually test but should get the idea across)

christianvuerings (approved):
Approving to unblock but would also be great if we can add a unit test to test the actually rotation + color for a couple of words. We can do that now since it's no longer random

packages/syntax-core/src/WordConfetti/WordConfetti.tsx
jliotta: Should I be individually memoizing the hash of each word? I memoize the entire result of `styledWords` so I think that should be sufficient

packages/syntax-core/src/WordConfetti/WordConfetti.test.tsx
```diff
@@ -26,4 +26,22 @@ describe("wordConfetti", () => {
     expect(screen.getByText("test")).toBeInTheDocument();
     expect(screen.getByText("word")).toBeInTheDocument();
   });
+
+  it("returns a consistent color and tilt for words", () => {
```
jliotta, line 30: @christianvuerings What's the best way to test rotation and background color? I had to add a data-testid to `WordConfetto` but right now it's just the text itself which isn't unique.

Also not sure if the uses of `toMatchInlineSnapshot` are how we want to do this

packages/syntax-core/src/WordConfetti/WordConfetti.tsx
christianvuerings: Yes, should be sufficient

packages/syntax-core/src/WordConfetti/WordConfetti.tsx
```diff
@@ -50,6 +50,7 @@ const WordConfetto = ({
 }): ReactElement => {
   return (
     <Box
+      data-testid={text}
```
christianvuerings, line 53: @jliotta we should avoid dynamic `data-testid`'s like this

christianvuerings:
@jliotta created a follow up at https://github.com/Cambly/syntax/pull/559

packages/syntax-core/src/WordConfetti/WordConfetti.test.tsx
christianvuerings: @jliotta see https://github.com/Cambly/syntax/pull/559 for an example. Yes, it's possible to use `toMatchInlineSnapshot` but we should then avoid adding the padding values as well

packages/syntax-core/src/WordConfetti/WordConfetti.tsx
christianvuerings: @jliotta see https://github.com/Cambly/syntax/pull/559 for a solution - we still use the parent test id + the index

packages/syntax-core/src/WordConfetti/WordConfetti.tsx
```diff
@@ -108,9 +120,8 @@ const WordConfetti = forwardRef<HTMLDivElement, WordConfettiProps>(
       () =>
         words.map((word) => ({
           text: word,
-          backgroundColor:
-            themeBackgroundColors[theme][Math.floor(Math.random() * 3)],
-          rotation: degreeOfTiltOptions[Math.floor(Math.random() * 9)],
+          backgroundColor: themeBackgroundColors[theme][hashCode(word) % 3],
```
christianvuerings, line 123: @jliotta thinking about these numbers `3` and `5`, we should probably base those on the length of options instead. See https://github.com/Cambly/syntax/pull/559 for a solution

PR #564, opened 2024-10-04
ZIndex: base implementation + story

somethiiing:
Z-Index implementation based off of [Outlayers UI Layer System](https://medium.com/@bernardocardoso/outsystems-ui-layer-system-managing-z-index-at-scale-68dca9e543de)

Basically has 6 "groups" of layers. and a `localLayer` prop to adjust between the layers if necessary

image

### DIFF
--- a/apps/storybook/stories/ZIndex.stories.mdx
+++ b/apps/storybook/stories/ZIndex.stories.mdx
@@ -1,0 +1,30 @@
+import { Meta } from "@storybook/addon-docs";
+import ZIndex from "./ZIndex";
+
+<Meta title="z-Index" />
+
+# z-Index
+
+[getZIndex](https://cambly-syntax.vercel.app/?path=/docs/z-index--docs) is a utility function that returns a z-index value based on the layer and local layer.
+
+## Parameters
+
+- `layer`: The layer to get the z-index for.
+
+  - Default: `base`
+  - Options:
+    - `base`: 0 // base layer
+    - `menu`: 10 // example: dropdown menus
+    - `sticky`: 20 // example: sticky headers
+    - `fixed`: 30 // example: fixed position elements
+    - `modal`: 40 // example: modals
+    - `popup`: 50 // example: tooltips or popovers
+
+- `localLayer`: An optional parameter to manually tweak the z-index within each layer.
+  - Usage:
+    - getZIndex recognizes that there are occasional needs to have multiple things stacked in the same layer.
+    - For example, two modals existing in the same layer. In these occasions, you can pass a localLayer to the function to manually tweak the z-index within each layer.
+  - Default: `0`
+  - Options: `-3` | `-2` | `-1` | `0` | `1` | `2` | `3`
+
+<ZIndex />
--- a/apps/storybook/stories/ZIndex.tsx
+++ b/apps/storybook/stories/ZIndex.tsx
@@ -1,0 +1,146 @@
+import React, { useState } from "react";
+import Box from "../../../packages/syntax-core/src/Box/Box";
+import Button from "../../../packages/syntax-core/src/Button/Button";
+import Typography from "../../../packages/syntax-core/src/Typography/Typography";
+import Modal from "../../../packages/syntax-core/src/Modal/Modal";
+import getZIndex from "../../../packages/syntax-core/src/getZIndex";
+
+export default function ZIndex() {
+  const [currentZIndex, setCurrentZIndex] = useState(-10);
+  const [openModal, setOpenModal] = useState(false);
+
+  return (
+    <Box display="flex" direction="column" gap={4}>
+      <Box>
+        <Typography size={400} weight="medium">
+          Current Z-Index: {currentZIndex}
+        </Typography>
+        <Box display="flex" gap={4}>
+          <Button
+            text="Increase Z-Index Layer"
+            onClick={() => setCurrentZIndex(currentZIndex + 10)}
+          />
+          <Button
+            text="Decrease Z-Index Layer"
+            onClick={() => setCurrentZIndex(currentZIndex - 10)}
+          />
+          <Button text="Open Modal" onClick={() => setOpenModal(true)} />
+        </Box>
+      </Box>
+      <Box
+        height={1000}
+        width="100%"
+        backgroundColor="gray100"
+        display="flex"
+        direction="column"
+        gap={4}
+        dangerouslySetInlineStyle={{
+          __style: {
+            border: "1px solid black",
+          },
+        }}
+      >
+        <Box position="relative">
+          <Box
+            position="absolute"
+            width="100%"
+            height={1000}
+            dangerouslySetInlineStyle={{
+              __style: {
+                backgroundColor: "darkgray",
+                zIndex: currentZIndex,
+              },
+            }}
+          />
+        </Box>
+        <Box
+          width="100%"
+          height={64}
+          display="flex"
+          justifyContent="center"
+          alignItems="center"
+          dangerouslySetInlineStyle={{
+            __style: {
+              backgroundColor: "lightblue",
+              zIndex: getZIndex("sticky"),
+            },
+          }}
+        >
+          Theoretical sticky navbar (sticky layer)
+        </Box>
+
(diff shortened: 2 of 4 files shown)


apps/storybook/stories/ZIndex.stories.mdx
```diff
@@ -0,0 +1,30 @@
+import { Meta } from "@storybook/addon-docs";
+import ZIndex from "./ZIndex";
+
+<Meta title="z-Index" />
```
somethiiing, line 4: autodocs unfortunately dont work here because it's a function and not an actual component. so i manually created docs on how to use this function

apps/storybook/stories/ZIndex.stories.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+# z-Index
+
+[getZIndex](https://cambly-syntax.vercel.app/?path=/docs/z-index--docs) is a utility function that returns a z-index value based on the layer and local layer.
```
christianvuerings, line 8: So it also works locally or when we 1 day change our domain to something else:
```suggestion
[getZIndex](?path=/docs/z-index--docs) is a utility function that returns a z-index value based on the layer and local layer.
```

apps/storybook/stories/ZIndex.stories.mdx
```diff
@@ -0,0 +1,35 @@
...
+```
+
+## Parameters
+
+- `layer`: The layer to get the z-index for.
+
+  - Default: `base`
+  - Options:
+    - `base`: 0 // base layer
+    - `menu`: 10 // example: dropdown menus
```
christianvuerings, line 22: @somethiiing should we remove `example: ` from every line?

apps/storybook/stories/ZIndex.tsx
```diff
@@ -0,0 +1,146 @@
...
+                backgroundColor: "#FFCC99",
+              },
+            }}
+          >
+            <Typography>Theoretical modal content (modal layer)</Typography>
+            <Typography>
+              tiny bug on this page, backdrop doesnt show because of no color
+              tokens but this is just used to demonstrate the layers!
+            </Typography>
+          </Box>
```
christianvuerings, line 141: @somethiiing Could we just add `import "@cambly/syntax-design-tokens/dist/css/variables.css";` to fix this bug?

### Before
Screenshot 2024-10-04 at 6 25 34 AM

### After
Screenshot 2024-10-04 at 6 25 27 AM

packages/syntax-core/src/getZIndex.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * [getZIndex](https://cambly-syntax.vercel.app/?path=/docs/ZIndex--docs) is a utility function that returns a z-index value based on the layer and local layer.
```
christianvuerings, line 2: Note that in this case, we do need to leave the URL as is since otherwise it won't work with TSDoc

packages/syntax-core/src/getZIndex.ts
```diff
@@ -0,0 +1,42 @@
...
+  localLayer: -3 | -2 | -1 | 0 | 1 | 2 | 3 = 0,
+): number {
+  const layerIndex = {
+    base: 0,
+    menu: 10,
+    sticky: 20,
+    fixed: 30,
+    modal: 40,
+    popup: 50,
+  };
```
christianvuerings, line 39:
```suggestion
  } as const;
```

apps/storybook/stories/ZIndex.tsx
```diff
@@ -0,0 +1,146 @@
...
+        }}
+      >
+        <Box position="relative">
+          <Box
+            position="absolute"
+            width="100%"
+            height={1000}
+            dangerouslySetInlineStyle={{
+              __style: {
+                backgroundColor: "darkgray",
```
christianvuerings, line 50: @somethiiing could we use the regular `backgroundColor` prop on `Box`?

christianvuerings (commented):
@somethiiing thanks for putting this together!

I like the overall directions but just had a couple of notes:

1) How are we thinking of enforcing this? In Gestalt, we did this by enforcing a certain type e.g. `Indexable` instead of `number`. See https://github.com/pinterest/gestalt/blob/d86286b7f74e51e24bf5ce67ac55e6a208205624/packages/gestalt/src/Box.tsx#L518 &  https://gestalt.pinterest.systems/web/zindex_classes 
1) Ideally we make the zIndex's explicitly depend on each other. E.g. how do we know whether we should pass a `localLayer` of `0`, `1`, `2` or `3`? That usually depends on another component 
1) Be sure to add `getZIndex` to the syntax core exports https://github.com/Cambly/syntax/blob/804cc971fd154f336a477e454c0e85b5ea423543/packages/syntax-core/src/index.tsx#L1-L31
1) Just commenting for now since it would be good for someone else to also take a deeper look

packages/syntax-core/src/getZIndex.ts
```diff
@@ -0,0 +1,42 @@
...
+  layer: "base" | "menu" | "sticky" | "fixed" | "modal" | "popup" = "base",
+  /**
+   * getZIndex recognizes that there are occasional needs to have multiple things stacked in the same layer.
+   * For example, two modals existing in the same layer. In these occasions, you can pass a localLayer to the function to manually tweak the z-index within each layer.
+   *
+   * @defaultValue `0`
+   */
+  localLayer: -3 | -2 | -1 | 0 | 1 | 2 | 3 = 0,
+): number {
+  const layerIndex = {
```
juliar, line 32: How well do these particular layers correspond to how we are using z-indexes in the app right now? I think it would be helpful to audit where we currently need z-indexes and for what types of components.

packages/syntax-core/src/getZIndex.ts
```diff
@@ -0,0 +1,42 @@
...
+   * @defaultValue `base`
+   */
+  layer: "base" | "menu" | "sticky" | "fixed" | "modal" | "popup" = "base",
+  /**
+   * getZIndex recognizes that there are occasional needs to have multiple things stacked in the same layer.
+   * For example, two modals existing in the same layer. In these occasions, you can pass a localLayer to the function to manually tweak the z-index within each layer.
+   *
+   * @defaultValue `0`
+   */
+  localLayer: -3 | -2 | -1 | 0 | 1 | 2 | 3 = 0,
```
juliar, line 30: How will folks know which localLayer to choose?

packages/syntax-core/src/getZIndex.ts
```diff
@@ -0,0 +1,42 @@
...
+): number {
+  const layerIndex = {
+    base: 0,
+    menu: 10,
```
juliar, line 34: Having comments/explanations for each of these layers will be really helpful, so people know which one is appropriate for their use case. Just from the names, I'm not sure when to use "base", "sticky", or "fixed", and when I should use "modal" vs "popup".

packages/syntax-core/src/getZIndex.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * [getZIndex](?path=/docs/ZIndex--docs) is a utility function that returns a z-index value based on the layer and local layer.
+ *
+ * Usage:
+ * ```
+ * <Navbar zIndex={getZIndex("sticky")} />
+ * <Modal zIndex={getZIndex("modal")} />
+ * ```
+ */
+export default function getZIndex(
```
juliar, line 10: How are you thinking the migration process would work, to move from our current z-indexes to these? Would we need to migrate all z-indexes at once?

packages/syntax-core/src/getZIndex.ts
juliar: Keep in mind that some z-indexes might not even be explicitly set in our code. For example, it seems that the MUI Dialog component has a z-index of 1300 (see [previous discussion](https://github.com/Cambly/Cambly-Frontend/pull/11895#discussion_r1750587506)), and it appears we are still using that in some places in our code.

juliar (commented):
Thanks for making this happen Wilson! This is a great start. I haven't looked deeply at your code, but added a few high level thoughts/questions to start.

PR #574, opened 2024-10-22
Tooltip: update api + add zIndex as optional prop

somethiiing:
Update tooltip to have easier to use api

+ add zIndex as optional prop

### DIFF
--- a/.changeset/nervous-brooms-raise.md
+++ b/.changeset/nervous-brooms-raise.md
@@ -1,0 +1,5 @@
+---
+"@cambly/syntax-floating-components": major
+---
+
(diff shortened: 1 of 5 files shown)


nathan-ong (approved):

williamnewman94 (commented):
LGTM!! Thank you Nathan!!

JFChoe:
If we don't need the content to be anything more than a string, should we consider deprecating `syntax-floating-components` and just use the `Tooltip` from `syntax-core`? 

Besides support for complex interactions, I think the only other advantage Floating-UI provides is the dynamic positioning to avoid collision with the viewport

PR #577, opened 2024-10-22
Allow option to preserve whitespace for Typography content

dasfrosty:
## Changes

Add support to `Typography` component for setting white-space CSS property.

### For example:
Screenshot 2024-10-22 at 4 06 41 PM

### DIFF
--- a/packages/syntax-core/src/Typography/Typography.tsx
+++ b/packages/syntax-core/src/Typography/Typography.tsx
@@ -127,6 +127,20 @@ const Typography = forwardRef<
      * @defaultValue "regular"
      */
     weight?: "regular" | "medium" | "semiBold" | "bold";
+    /**
+     * Controls how whitespace within the element is handled.
+     *
+     * See: https://developer.mozilla.org/en-US/docs/Web/CSS/white-space
+     *
+     * @defaultValue "inherit"
+     */
+    whiteSpace?:
+      | "inherit"
+      | "normal"
+      | "nowrap"
+      | "pre"
+      | "preLine"
+      | "preWrap";
   }
 >(function Typography(
   {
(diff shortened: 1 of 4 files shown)


packages/syntax-core/src/Typography/Typography.tsx
```diff
@@ -127,6 +127,20 @@ const Typography = forwardRef<
...
      */
     weight?: "regular" | "medium" | "semiBold" | "bold";
+    /**
+     * Controls how whitespace within the element is handled.
+     *
+     * See: https://developer.mozilla.org/en-US/docs/Web/CSS/white-space
+     *
+     * @defaultValue "inherit"
+     */
+    whiteSpace?:
```
somethiiing, line 137: hmm. generally, we've limited alot of these props to only the ones we need. ive inherited christian's rather be more strict and then if we need the others ones we'll add it later

how often do you think we'll need all of these? 
i imagine just inherit, normal, nowrap and pre will be fine

dasfrosty: For the current use case we want `pre-line` so that we can preserve line breaks but still wrap.
Screenshot 2024-10-22 at 5 15 10 PM

I'll limit it to `inherit`, `normal`, `nowrap` and `pre-line` for now.

Unless you think there's a use case for `pre` as well:

Screenshot 2024-10-22 at 5 07 51 PM

somethiiing: just the ones you mentioned are fine. thanks!

dasfrosty: > just the ones you mentioned are fine. thanks!

Done!

somethiiing (approved):

PR #590, opened 2024-10-29
[NOJIRA] Add option popoverZIndex prop to RichSelectList

williamnewman94:
Black Friday is about to begin and our tutor search page is broken:

Screenshot 2024-10-29 at 12 33 30 PM

Tested to see that it forwarded the z-index to the popover properly:

Screenshot 2024-10-29 at 3 45 55 PM

Needed a quick fix because black friday is beginning today. If you have a more robust fix in mind let me know.

### DIFF
--- a/packages/syntax-core/src/RichSelect/RichSelectList.tsx
+++ b/packages/syntax-core/src/RichSelect/RichSelectList.tsx
@@ -72,7 +72,7 @@ export type RichSelectListProps = RichSelectBoxProps & {
   /**
    * Z-index of the popover
    */
-  popoverZIndex?: number;
+  zIndex?: number;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


somethiiing:
dont forget to run `pnpm run changeset`

packages/syntax-core/src/RichSelect/RichSelectList.tsx
```diff
@@ -69,6 +69,10 @@ export type RichSelectListProps = RichSelectBoxProps & {
    * @defaultValue white
    */
   color?: "white" | "clear";
+  /**
+   * Z-index of the popover
+   */
+  popoverZIndex?: number;
```
somethiiing, line 75: can you just call this zIndex?

somethiiing (approved):
please run changeset to ensure the npm package gets updated + update prop name.

would be really great to update the story with the new prop as well. (you dont have to add a custom story highlighting, but would be great to see it in the prop list)

williamnewman94:
> please run changeset to ensure the npm package gets updated + update prop name.
> 
> would be really great to update the story with the new prop as well. (you dont have to add a custom story highlighting, but would be great to see it in the prop list)

Woops sorry forgot about the changeset!! Added a story as well though I don't think that it is hooked up properly because the change I make to the only param I can change (disabled) does not affect the example.

PR #592, opened 2024-10-31
[SYNTAX] Create LinkTapArea

ChessyHsu:
[Jira ticket](https://cambly.atlassian.net/browse/WEB-6?atlOrigin=eyJpIjoiYWU0NTQ3YjllOTMyNDY3OWE1M2U5YjQxYWQ2NzUwYTEiLCJwIjoiaiJ9)

Create LinkTapArea by replacing div in TapArea with a tag. Now LinkTapArea behaves like a tag, it provides accessibility for users who rely on assistive technologies and shows url when you hover over it on the bottom left corner.

Changes:
- add LinkTapArea
- add LinkTapArea to storybook
- add unit tests for LinkTapArea
- run pnpm changeset

Screenshot 2024-10-31 at 3 18 50 PM

PR #595, opened 2024-11-06
[SYNTAX] Make LinkTapArea a link

ChessyHsu:
Followup on creating a LinkTapArea https://github.com/Cambly/syntax/pull/592

1. Change LinkTapArea from a button to a link: 
TapArea is a button made by div with onClick taking in synthetic events while LinkTapArea should naturally be a link and thus its onClick takes in mouse events.
2. Remove the disable attribute: since clicking on a link trigger a redirection to other page, it should always be available and does not need to wait for actions to complete.

 
image


nathan-ong:
@ChessyHsu , can you include screenshots?

jliotta (approved):

PR #596, opened 2024-11-07
Typography/Korean lang no word break

somethiiing:
add word-break: keep-all for korean

### DIFF
--- a/packages/syntax-core/src/Typography/Typography.module.css
+++ b/packages/syntax-core/src/Typography/Typography.module.css
@@ -1,4 +1,5 @@
 .typography {
+  word-break: keep-all;
   line-height: initial;
   margin: 0;
 }
@@ -52,6 +53,10 @@
 * Simplified Chinese: ❌
 * Traditional Chinese: ❌
 */
+
+:lang(ko) {
+  word-break: keep-all;
+}
 
 /* stylelint-disable-next-line selector-max-compound-selectors -- only render serif font for certain languages */
 :lang(en) .serif,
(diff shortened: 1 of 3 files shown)


zprong (approved):
I see this would also be the case for Chinese and Japanese -- can we update this to include `ja` and `zh`?

fransiskusx (approved):
lgtm

zprong:
source: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#keep-all

somethiiing:
> source: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#keep-all

for zh and ja, i think we need a different strategy. it doenst work entirely as intended when i tested it

packages/syntax-core/src/Typography/Typography.module.css
```diff
@@ -229,3 +229,8 @@
 .preLine {
   white-space: pre-line;
 }
+
+:lang(ko),
+.typography {
```
christianvuerings, line 234: @somethiiing Did you check whether this is correct? I noticed that all text now has `word-break: keep-all`, not just Korean.

```suggestion
/* stylelint-disable-next-line selector-max-compound-selectors -- https://github.com/w3c/csswg-drafts/issues/4285 */
:lang(ko) .typography {
  word-break: keep-all;
}
```

Also, I would move these lines to where we have the other language specific overrides:
https://github.com/Cambly/syntax/blob/76d59226915367c84f9ccdf0497de9911c146398/packages/syntax-core/src/Typography/Typography.module.css#L57-L58

image

PR #606, opened 2025-02-10
[NOJIRA] Syntax Modal Styling Changes

ellenxwchen:
Multiple changes for Syntax modal:
1. Change title to 400 semibold + not serif style
2. Change X icon button to size SM
3. Put title and X icon in the same div so they never overlap
4. Add a divider between title and content
5. Remove divider when image is present
6. Bumps zIndex on stories to stop the "Show code" from being on top of the modal
7. Changes the padding from 6 to 4

https://github.com/user-attachments/assets/5525b878-0ee1-4a3d-9eae-4f734a39a651

https://github.com/user-attachments/assets/a79d4767-a5b4-4386-a22b-0570197e1486

### DIFF
--- a/packages/syntax-core/src/Modal/Modal.tsx
+++ b/packages/syntax-core/src/Modal/Modal.tsx
@@ -210,7 +210,6 @@ export default function Modal({
                   overflowY="auto"
                   paddingX={4}
                   paddingY={image ? 0 : 4}
-                  marginBottom={image ? 4 : 0}
                 >
                   {children}
                 </Box>
--- a/packages/syntax-core/src/Modal/Modal.tsx
+++ b/packages/syntax-core/src/Modal/Modal.tsx
@@ -8,6 +8,7 @@ import StopScroll from "./StopScroll";
 import Layer from "./Layer";
 import styles from "./Modal.module.css";
 import IconButton from "../IconButton/IconButton";
+import Divider from "../Divider/Divider";
 
 function XIcon({ className }: { className?: string }) {
   return (
@@ -194,6 +195,17 @@ export default function Modal({
                     />
                   )}
                 </Box>
+                {!image && (
+                  <Box
+                    display="flex"
+                    direction="column"
+                    width="100%"
+                    paddingX={4}
+                    marginBottom={4}
+                  >
+                    <Divider />
+                  </Box>
+                )}
                 <Box
                   height="100%"
                   overflowY="auto"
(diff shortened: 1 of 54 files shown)


packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -152,34 +153,64 @@ export default function Modal({
...
-                  <Heading as="h1" size={500} fontStyle="serif">
+                <Box
+                  padding={4}
+                  width="100%"
+                  direction="row"
+                  display="flex"
+                  gap={1}
+                  justifyContent="between"
+                >
+                  <Heading as="h1" size={400}>
```
emmanuelgamboa09, line 184: the serif removal was intentional correct?.

packages/syntax-core/src/Modal/Modal.tsx
```diff
@@ -152,34 +153,64 @@ export default function Modal({
...
+                  </Box>
+                )}
                 <Box
                   height="100%"
                   overflowY="auto"
-                  paddingX={6}
-                  marginBottom={footer ? 0 : 6}
+                  paddingX={4}
+                  paddingY={image ? 0 : 4}
+                  marginBottom={image ? 4 : 0}
```
emmanuelgamboa09, line 213: huh is marginBottom just because box doesn't have paddingBottom or something. Sort of interesting that we alternate between the two, but I imagine thats just a work around the constraints of box.

emmanuelgamboa09 (approved):

packages/syntax-core/src/Modal/Modal.tsx
ellenxwchen: Yea, essentially we don't want paddingY={4} on the image because it's for the `<Divider/>` (so it's |Title|4|Divider|4|Content|4|), but there's no `<Divider/>` on image modal designs, so this would give us padding={8} between the Title and the Content which we don't want.

I can't set the marginBottom on the Divider and I can't do `marginTop={image ? 0 : 4} marginBottom={4}` here, because we want the scrollable content to touch the Divider like this 
Screenshot 2025-02-10 at 4 42 54 PM

packages/syntax-core/src/Modal/Modal.tsx
ellenxwchen: Yes, correct, this removal was intentional

PR #608, opened 2025-02-25
[NOJIRA] Badge: Add silver colour, change text size to 0

ellenxwchen:
Figma: https://www.figma.com/design/7wAM5Us9AnFrYHyLV1YCEs/Premium-Tier?node-id=2023-6014&m=dev
Badge:
Adds silver colour for premium.
Text size should have been 0 instead of 100 here.

Screenshot 2025-02-25 at 3 18 54 PM

Screenshot 2025-02-25 at 3 18 59 PM

### DIFF
--- a/packages/syntax-core/src/Badge/Badge.tsx
+++ b/packages/syntax-core/src/Badge/Badge.tsx
@@ -73,7 +73,7 @@ const Badge = ({
       paddingX={2}
       paddingY={1}
       rounding={"sm"}
-      backgroundColor={color === "silver" ? undefined : color}
+      backgroundColor={color !== "silver" ? color : undefined}
       alignItems="center"
       justifyContent="center"
       minHeight={24}
--- a/packages/syntax-core/src/Badge/Badge.tsx
+++ b/packages/syntax-core/src/Badge/Badge.tsx
@@ -50,21 +50,8 @@ const backgroundColorForColor = (
   switch (color) {
     case "silver":
       return undefined;
-    case "gray370":
-    case "destructive300":
-    case "orange":
-    case "tan":
-    case "success300":
-    case "sky":
-    case "thistle":
-    case "pink":
-    case "lilac":
-    case "cream":
-    case "yellow700":
-    case "gray870":
-      return color;
     default:
-      return undefined;
+      return color;
   }
 };
 
@@ -79,18 +66,6 @@ const inlineStylesForColor = (
           linear-gradient(83.45deg, #A9A9A9 2.57%, #E5E2E2 61.77%, #6E6E6E 100.3%) border-box",
         border: "1px solid transparent",
       };
-    case "gray370":
-    case "destructive300":
-    case "orange":
-    case "tan":
-    case "success300":
-    case "sky":
-    case "thistle":
-    case "pink":
-    case "lilac":
-    case "cream":
-    case "yellow700":
-    case "gray870":
     default:
       return {};
   }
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -71,21 +73,33 @@ const Badge = ({
       paddingX={2}
       paddingY={1}
       rounding={"sm"}
-      backgroundColor={color}
+      backgroundColor={color !== "silver" ? color : undefined}
```
emmanuelgamboa09, line 76: Should this be called premium or something. its not really silver right. More of a custom silver. But not sure myself.

emmanuelgamboa09, line 76: can we turn this into a function to get the color. Something similiar to textColorForBackgroundColor type of deal

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -71,21 +73,33 @@ const Badge = ({
...
+          ? {
+              __style: {
+                background:
+                  "linear-gradient(85deg, #CECECE -8.89%, #EEECEC 38.35%, #FFF 49.64%, #E9E8E8 66.22%) padding-box, \
+                  linear-gradient(83.45deg, #A9A9A9 2.57%, #E5E2E2 61.77%, #6E6E6E 100.3%) border-box",
+                border: "1px solid transparent",
+              },
+            }
+          : { __style: {} }
+      }
```
emmanuelgamboa09, line 91: yea if we can also that that here if possible. Feel free to give me your thoughts ont his.

packages/syntax-core/src/Badge/Badge.tsx
ellenxwchen: That's a really good point tysm!

ellenxwchen: That's a good point - I think since the rest of the colour options are closest to colour names, I align more to keeping "silver" since we might reuse this for non-premium work. Mostly also because the "premium" naming is still up in the air now, I'd rather keep it more general for now, and I can chat with Joey about it later.

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -42,6 +44,58 @@ const textColorForBackgroundColor = (
   }
 };
 
+const backgroundColorForColor = (
+  color: BadgeColor,
+): Exclude<BadgeColor, "silver"> | undefined => {
+  switch (color) {
+    case "silver":
```
emmanuelgamboa09, line 51: is this necessary here. Doesn't default take care of the undefined value.

emmanuelgamboa09: Could we do something like
```
switch (color) {
   case "silver":
       return undefined;
   default:
      return color
}
```

packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -42,6 +44,58 @@ const textColorForBackgroundColor = (
...
+    case "orange":
+    case "tan":
+    case "success300":
+    case "sky":
+    case "thistle":
+    case "pink":
+    case "lilac":
+    case "cream":
+    case "yellow700":
+    case "gray870":
```
emmanuelgamboa09, line 93: same here I don't think we need to add these cases. I think it can remove all these cases but if theres a reason we actually do this let me know. Not sure myself.

packages/syntax-core/src/Badge/Badge.tsx
ellenxwchen: No you're right, that looks much cleaner

emmanuelgamboa09 (approved):

PR #609, opened 2025-02-26
Update Chip to support end icon

emmanuelgamboa09:
Chip:
Add endIcon to chip
Updated icon to be startIcon

### DIFF
--- a/packages/syntax-core/src/Chip/Chip.tsx
+++ b/packages/syntax-core/src/Chip/Chip.tsx
@@ -1,7 +1,6 @@
 import React, { forwardRef, useMemo, type ComponentProps } from "react";
 import classnames from "classnames";
 import Typography from "../Typography/Typography";
-import Box from "../Box/Box";
 import styles from "./Chip.module.css";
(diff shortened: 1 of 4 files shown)


packages/syntax-core/src/Chip/Chip.tsx
```diff
@@ -139,18 +145,23 @@ const Chip = forwardRef<HTMLButtonElement, ChipProps>(
...
+          {text}
+        </Typography>
+        {EndIcon && (
+          <EndIcon
             className={classnames(iconStyles, materialIconSize[size])}
             color={color}
             size={internalIconSize[size]}
           />
         )}
-        <Box paddingX={Icon ? 1 : 0}>
```
emmanuelgamboa09, line 149: This can be accomplished with gap instead of paddingX and this conditional value. Since gap will only apply when there are multiple components inside of the parent component. So in this case we don't care of there is a start or end icon. When they are added they automatically get the spaccing applied to them.

ellenxwchen (approved):

PR #612, opened 2025-02-27
Lock eslint config syntax packages

emmanuelgamboa09:
These packages weren't lock which was leading to lint failing within main. This goes in and locks eslint-config-next and eslint-config-turbo to the version we were on prior to it starting to fail.


ellenxwchen (approved):
Thank you

PR #627, opened 2025-06-24
New badge color, used by Pro badge

srini91:
blackandwhite - a black and white gradient

Example (ignore the yellow line; that was just a bad screenshot cut):
Screenshot 2025-06-23 at 6 18 10 PM

Screenshot 2025-06-23 at 6 18 17 PM

### DIFF
--- a/packages/syntax-core/src/Badge/Badge.tsx
+++ b/packages/syntax-core/src/Badge/Badge.tsx
@@ -18,6 +18,7 @@ const badgeColor = [
   "cream",
   "yellow700",
   "silver",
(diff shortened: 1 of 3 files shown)


packages/syntax-core/src/Badge/Badge.tsx
```diff
@@ -46,10 +47,12 @@ const textColorForBackgroundColor = (
```
ellenxwchen, line 47: I can't comment on it, but can we add `blackandwhite` to the switch statement for white in `textColorForBackgroundColor`?

PR #634, opened 2025-09-10
Fix: recover icon adapt dir

ASkyBig:
As we have finished modification of those icons in our frontend project, now we can merge it and upgrade version in frontend project.
Original PR: https://github.com/Cambly/syntax/pull/622
Reverted PR: https://github.com/Cambly/syntax/pull/629, we reverted the original PR because we didn't finished it before, then icons will be wrong direction on rtl document.

### DIFF
--- a/.changeset/nice-geese-greet.md
+++ b/.changeset/nice-geese-greet.md
@@ -1,0 +1,6 @@
+---
+"@cambly/syntax-icons": patch
+"@cambly/syntax-core": patch
+---
+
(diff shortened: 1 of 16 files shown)


ellenxwchen (approved):
Not sure why the deployment is failing, I don't have access to Vercel, but can you try forcing another deployment in case it's not a one-off deployment issue?

ericsandine:
@ASkyBig I think we can merge this one now?

PR #671, opened 2026-04-23
Add light/dark toggle for Storybook UI chrome

zomerfeld:
## Summary
- Adds `@vueless/storybook-dark-mode` to the Storybook app so users can toggle the Storybook manager UI between light and dark from the toolbar.
- The toggle only affects **Storybook's chrome** (sidebar, toolbar, docs panels). The story canvas and the components being previewed are deliberately untouched — `stylePreview` is omitted.
- Splits `syntaxTheme.ts` into a `darkTheme` / `lightTheme` pair (shared brand metadata) and moves theme selection out of `manager.ts` and into `preview.tsx`'s `darkMode` parameter, so the addon controls both themes.

## Why
- We picked `@vueless/storybook-dark-mode` over:
  - `@storybook/addon-themes` — it's the official addon but only themes the canvas, not the Storybook UI, which is what we actually wanted here.
  - `hipstersmoothie/storybook-dark-mode` — the original community addon; compatibility with Storybook 10 has lagged.
  - `@vueless/storybook-dark-mode` is a maintained fork with the same API, explicitly tracking Storybook 9/10. We're on 10.3.5.

## Scope — intentionally narrow
- No component styling changes.
- `ThemeProvider` continues to own the Syntax runtime theme for the stories themselves.
- Our `ThemeProvider` / design-token side is unaffected.

## Future extensions
The addon already exposes the primitives we'd need if we later want to theme the canvas too:
- Flip `stylePreview: true` in `preview.tsx` to have the addon apply `dark` / `light` classes to the preview iframe. We could pair that with `classTarget` to land the class where Syntax's token CSS expects it.
- Use the exported `useDarkMode()` hook (or the `DARK_MODE_EVENT_NAME` channel event) inside a preview decorator to drive `ThemeProvider`'s theme prop from the same toggle — so the Storybook chrome and Syntax components switch together.
- Wire a matching light-mode token set into `syntax-design-tokens` once that's in scope, then the two above stop being "future work" and become a one-line config change.

## Test plan
- [ ] `pnpm start` — Storybook boots without errors.
- [ ] Toolbar shows the sun/moon toggle; clicking it swaps Storybook's chrome between the `lightTheme` and `darkTheme`.
- [ ] Preference persists across reloads (localStorage) and respects OS \`prefers-color-scheme\` on first load.
- [ ] Story canvas / rendered components stay visually identical in both modes.
- [ ] Docs pages render correctly in both modes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/storybook/.storybook/main.ts
+++ b/apps/storybook/.storybook/main.ts
@@ -19,6 +19,7 @@ const config: StorybookConfig = {
     getAbsolutePath("storybook-addon-rtl"),
     getAbsolutePath("@storybook/addon-docs"),
     getAbsolutePath("@storybook/addon-mcp"),
+    getAbsolutePath("@vueless/storybook-dark-mode"),
   ],
 
   staticDirs: ["../public"],
--- a/apps/storybook/.storybook/manager.ts
+++ b/apps/storybook/.storybook/manager.ts
@@ -1,11 +1,4 @@
-import { addons } from "storybook/manager-api";
-import syntaxTheme from "./syntaxTheme";
-
 const link = document.createElement("link");
 link.setAttribute("rel", "shortcut icon");
 link.setAttribute("href", "favicon.ico");
 document.head.appendChild(link);
-
-addons.setConfig({
-  theme: syntaxTheme,
-});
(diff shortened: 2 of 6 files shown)


philrl:
@greptileai Review this please

philrl (approved):

